PR #1, opened 2022-03-07
Basic project structure + Build scripts

mariomac:
This PR includes:
* Basic project structure
* Basic test+build scripts (Makefile and Dockerfile)
* A proof-of-concept code that is going to serve as starting point for the actual agent. Don't care too much about it as it will be reworked completely but it was useful to "test" the build scripts for BPF code generation.


mariomac:
putting on hold, as I'd like to do a later investigation about wrapping https://github.com/libbpf/libbpf-bootstrap, which seems to work pretty well, in Go.

jpinsonneau (approved):
LGTM

bpf/flows.c
```diff
@@ -0,0 +1,79 @@
...
+           (((u32)(x)&0x00ff0000U) >> 8) |  \
+           (((u32)(x)&0x0000ff00U) << 8) |  \
+           (((u32)(x)&0x000000ffU) << 24)))
+
+struct {
+    __uint(type, BPF_MAP_TYPE_RINGBUF);
+    __uint(max_entries, 1 << 24);
+} flows SEC(".maps");
+
+struct egress_t {
```
OlivierCazade, line 24: My understanding is that this struct and the go struct RawStats have to be equivalent.

Anyway we could generate one from the other?

OlivierCazade (approved):
Thanks, this is great to see the first step here!

LGTM, my comments are more about the code and this is going to change since this is a POC

bpf/headers/update.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+# Version of libbpf to fetch headers from
+LIBBPF_VERSION=0.7.0
```
jotak, line 4: How can we make sure we fetch the correct version? libbpf comes pre-installed, or do we install it?

mariomac: Libbpf is bundled with the Cilium library. We just need the headers for compiling the C code. I'll check if there is a way that we can fetch the Cilium's libbpf version. The destination machine does not need to install any extra library, as libbpf is only used to build the eBPF VM bytecode.

bpf/flows.c
mariomac: Unfortunately, I don't know any way (unless we create some kind of parser, which it would be a nice side project)

PR #2, opened 2022-03-11
NETOBSERV-200: Basic flows' agent architecture

mariomac:
It provides a basic agent architecture that fetches very basic flow
information and prints it in standard output.

This PR does not want to provide complete Flow's information. It
will be done in a later PR.


OlivierCazade (approved):
LGTM

Side question not related to this PR, is there a particular reason to track bpf object files?

mariomac:
@OlivierCazade I usually prefer to put generated files into a `generated` folder structure (like we usually do in Java) and exclude it from Git. However, the functions generated inside the `bpf_bpfe*` files are private and must be in the same folder as the `tracer.go` file from us.

If you prefer, we can anyway include these files individually in the `.gitignore`.

PR #3, opened 2022-03-17
NETOBSERV-200: Basic eBPF agent that fetches flows' information

mariomac:
This agent just prints flows in the standard output and does not forward to any other place.

Here is an example of the fetched flow information (the structure will change when we serialize it to a binary encoding):

```json
{
  "Etype": 8,
  "Direction": "EGRESS",
  "DataLink": {
    "SrcMac": "08:00:27:23:e8:8a",
    "DstMac": "52:54:00:12:35:02"
  },
  "Network": {
    "SrcAddr": "10.0.2.15",
    "DstAddr": "10.0.2.2"
  },
  "Transport": {
    "SrcPort": 22,
    "DstPort": 51819,
    "Proto": "TCP"
  },
  "Bytes": 56320,
  "TimeFlowStart": "2022-03-17T07:42:12.414480074Z",
  "TimeFlowEnd": "2022-03-17T07:42:12.414849771Z",
  "Interface": "eth0",
  "Packets": 2
}
```


pkg/flow/record.go
```diff
@@ -99,9 +137,28 @@ func (b HumanBytes) String() string {
...
+}
+
+func (d Direction) MarshalJSON() ([]byte, error) {
+	switch d {
+	case 0:
+		return []byte(`"INGRESS"`), nil
+	case 1:
+		return []byte(`"EGRESS"`), nil
+	default:
+		return []byte(`"UNKNOWN"`), nil
```
jpinsonneau, line 155: This should be impossible right ?

In general, I feel it would be better to remove some fields from JSON if input values are invalid than sending a default value that will make the parsing different for every fields.

jpinsonneau (approved):

pkg/flow/record.go
mariomac: Good catch. I left it there just to compile. Anyway consider that this was just to get a "nicer" demonstration of the flow information in JSON, but this will be removed in the next task, as long as we move to protobuf or any other binary encoding.

PR #6, opened 2022-03-22
[NETOBSERV-202] GRPC+Protobuf submission

mariomac:
* GRPC+Protobuf basic client/server implementation as a submodule so the code can be shared with flowlogs-pipeline.
* Example collector code that is used as a performance evaluator + deployment files.


pkg/grpc/client.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+// ClientConnection wraps a gRPC+protobuf connection
+type ClientConnection struct {
+	client pbflow.CollectorClient
+	conn   *grpc.ClientConn
+}
+
+func ConnectClient(address string) (*ClientConnection, error) {
+	// TODO: allow configuring some options (keepalive, backoff...)
```
jotak, line 17: Maybe set a default Backoff / Timeout in the meantime, before implementing it as a  configuration ?

mariomac: There are actually default values:
```go
var Default[Config](https://pkg.go.dev/google.golang.org/grpc/backoff#Config) = Config{
	BaseDelay:  1.0 * [time](https://pkg.go.dev/time).[Second](https://pkg.go.dev/time#Second),
	Multiplier: 1.6,
	Jitter:     0.2,
	MaxDelay:   120 * [time](https://pkg.go.dev/time).[Second](https://pkg.go.dev/time#Second),
}
```

jotak: ah, nice! I remember having to debug tricky bugs in the past due to not having a timeout, but that wasn't grpc, glad this one has a default :)

jotak (approved):

PR #8, opened 2022-03-28
Fix some CI tasks

mariomac:
It excludes generated files from the coverage reports, so the percentages are more realistic.

It also fixes a typo in the push_image.yml github action.

It also fixes a flaky test.


Makefile
```diff
@@ -92,22 +92,22 @@ compile:
 .PHONY: test
 test:
 	@echo "### Testing code"
-	GOOS=$(GOOS) go test -mod vendor -a ./... -coverpkg=./... -coverprofile cover.out
+	GOOS=$(GOOS) go test -mod vendor -a ./... -coverpkg=./... -coverprofile cover.all.out
```
OlivierCazade, line 95: Did you try to adapt the coverpkg option here?

In theory we should be able to exclude unwanted files here and directly have a clean output.

OlivierCazade (approved):
LGTM since my comment is related to a mechanism not modified by this PR.

Makefile
mariomac: Yeah I tried, but `coverpkg` only allows you picking up a single package and not a bunch of packages, nor just excluding selected packages.

OlivierCazade: Ok, thanks!

PR #16, opened 2022-04-20
NETOBSERV-201: configure LogLevel and Sampling

mariomac:
It replaces the "Verbose" binary config option by "LogLevel", to match other Netobserv components.

It also adds configurable sampling rate. 

The best way to test sampling is to use grafana to count the number of processed packets.

In the following image you can see the packet count during 3 differentiated periods: no sampling (100% of packets processed), sampling==2 (50% of packets processed) and sampling == 4 (25% of packets processed).

image


jotak (approved):
lgtm

PR #19, opened 2022-04-29
NETOBSERV-281: dynamic registering/unregistering of network interfaces

mariomac:
Outbound traffic for new Pods was not properly fetched unless you restarted the agent after the creation of the pod.

The cause was that, for each new pod, a new network interface is created in its node, and the eBPF agent does not discovered it, since it only inspected the interfaces at its initialization.


docs/config.md
```diff
@@ -24,3 +24,12 @@ The following environment variables are available to configure the NetObserv eBF
...
   processing stages. Most probably you won't need to change this value.
+* `LISTEN_INTERFACES` (default: `watch`). Mechanism used by the agent to listen for added or removed
+  network interfaces. Accepted values are:
+  - `watch`: interfaces are traced immediately after they are created. This is
+    the recommended setting for most configurations.
+  - `poll`: recommended mostly as a fallback mechanism if `watch` misbehaves. It periodically 
+    queries the current network interfaces. The poll frequency is specified by the
+    `LISTEN_POLL_PERIOD` variable.
+* `LISTEN_POLL_PERIOD` (default: `10s`). When `LISTEN_INTERFACES` value is `poll`, this duration
+  string specifies the frequency in which the current network interfaces are polled.
```
jpinsonneau, line 35: We should expose that in the operator config

Makefile
```diff
@@ -33,10 +33,12 @@ PROTOC_ARTIFACTS := pkg/pbflow
 EXCLUDE_COVERAGE_FILES="(/cmd/)|(bpf_bpfe)|(/examples/)|(/pkg/pbflow/)"
 
 # Image building tool (docker / podman)
-ifeq (,$(shell which podman 2>/dev/null))
-OCI_BIN=docker
-else
-OCI_BIN=podman
+ifndef OCI_BIN
```
jpinsonneau, line 36: :+1:

jpinsonneau (approved):
Tested both `watch` & `poll` and works as expected :1st_place_medal: 
Good job @mariomac !

docs/config.md
mariomac: The operator config now allows setting un an `env` section where you can set the Agent environment variables. On a first glance I decided to keep them just as env vars, hidden from the operator spec, to simplify the NOO documentation and hide from the common user some performance-tweaking arguments that are aimed for some very specific configurations (I'm thinking in customer support).

PR #29, opened 2022-05-17
Fixed example deployments

mariomac:
Example deployments had some missing entries, and other configuration properties were updated.


jotak (approved):
/lgtm

PR #30, opened 2022-05-19
NETOBSERV-199: E2E Integration tests

mariomac:
This PR just includes an initial, basic version of a test where:

1- A K8s environment is deployed with the Agent, Flowlogs-Pipeline and Loki.
2- An HTTP service is deployed, and a client pod periodically submits HTTP requests.
3- We verify that Loki has stored complete flows for the requests/responses. This would mean that both the Agent and Flowlogs-Pipeline have done their work.

The new `e2e` folder is divided (by now) in two parts:

* A `cluster` folder that provides the common part of each test: the basic cluster configuration and deployment files, and helper classes to run/access the Kind cluster and Loki.
* A `basic` folder that provides a concrete test: the HTTP client/server deployment plus the actual test features.

Additionally, it allows downloading the whole cluster logs to try to debug failed tests, from the artifacts section in the UI:

image


e2e/basic/manifests/pods.yml
```diff
@@ -0,0 +1,45 @@
...
+metadata:
+  name: client
+spec:
+  containers:
+    - name: fedora
+      image: fedora:36
+      command:
+        - sh
+        - -c
+        - "echo 'start pinging from client pod...'; while true; do curl -v http://server ; sleep 1; done"
```
jpinsonneau, line 14: What about sending some data in curl command here ? 

So we will be able to compare bytes of flows we got
`curl -v -d 'SOME_JSON_DATA' -H "Content-Type: application/json" -X POST http://server`

jpinsonneau: there is also `ping -s http://server` as an alternative

jpinsonneau (approved):
Looks awesome @mariomac ! Is there a quick way to run that locally ?
Thanks

mariomac:
Thanks for the review @jpinsonneau . I'll address your comments into another PR.

To run the tests locally, you should just run `make tests-e2e` (docker needed).

PR #31, opened 2022-06-02
NETOBSERV-333: allow overriding default GRPC server options

mariomac:
We need to instrument HTTP calls in flowlogs-pipeline. This PR allows injecting interceptors in the GRPC Server.


jpinsonneau (approved):
LGTM

PR #35, opened 2022-06-20
Fix FLP in e2e tests

mariomac:
Some breaking changes were added to FLP, making the e2e tests to fail. This PR fixes it.

### DIFF
--- a/e2e/cluster/base/03-flp.yml
+++ b/e2e/cluster/base/03-flp.yml
@@ -39,10 +39,8 @@ data:
     log-level: debug
     pipeline:
       - name: ingest
-      - name: decode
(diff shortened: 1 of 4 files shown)


jpinsonneau:
@mariomac can we also update `kafka` resources limits like:
```yaml
resources:
  limits:
    cpu: 1000m
    memory: 1024Mi
  requests:
    cpu: 250m
    memory: 256Mi
```

It solve my local issue where kafka never goes up since it's `OOMKilled`.
We may also consider setting limits on every deployment / pod.

jpinsonneau (changes requested):

PR #39, opened 2022-06-28
Performance optimization using eBPF aggregation

praveingk:
**Changed logic :**

1) Use Per-CPU hashmap in ebpf to aggregate flow records
2) Evict from userspace upon flow completion/timeout.

More logic details in [Here](https://github.com/netobserv/netobserv-ebpf-agent/blob/ebpf-opt/bpf/README.md)
Performance benefits in [Here](https://github.com/netobserv/netobserv-ebpf-agent/blob/ebpf-opt/docs/measurements.pptx)

Ran e2e tests and passed.


mariomac:
Adding @jpinsonneau to the loop, as owner of the eBPF improvement's epic

bpf/flow.h
```diff
@@ -5,11 +5,23 @@
 #define TC_ACT_SHOT 2
 #define IP_MAX_LEN 16
 
+// TODO : Explore if this can be programmed from go launcher
+// Total mem consumption would be (1000*)
+#define INGRESS_MAX_ENTRIES 1000
+#define EGRESS_MAX_ENTRIES  1000
```
mariomac, line 11: Yes, this can be configured from the Go launcher. As an example, the existing sampling configuration value:

1. [Define volatile const variables](https://github.com/netobserv/netobserv-ebpf-agent/blob/main/bpf/flows.c#L30)
2. [Use the RewriteConstants method from Go](https://github.com/netobserv/netobserv-ebpf-agent/blob/main/pkg/ebpf/tracer.go#L60-L62)

bpf/flows.c
```diff
@@ -65,89 +116,177 @@ static inline int fill_iphdr(struct iphdr *ip, void *data_end, struct flow *flow
...
         }
     } break;
     default:
         break;
     }
     return SUBMIT;
 }
-// sets flow fields from Ethernet header information
-static inline int fill_ethhdr(struct ethhdr *eth, void *data_end, struct flow *flow) {
+// // sets flow fields from Ethernet header information
```
mariomac, line 155:
```suggestion
// sets flow fields from Ethernet header information
```

bpf/flows.c
```diff
@@ -65,89 +116,177 @@ static inline int fill_iphdr(struct iphdr *ip, void *data_end, struct flow *flow
...
+        flow_id->protocol = 0;
+        flow_id->src_port = 0;
+        flow_id->dst_port = 0;
     }
     return SUBMIT;
 }
 
-// parses flow information for a given direction (ingress/egress)
-static inline int flow_parse(struct __sk_buff *skb, u8 direction) {
+static inline void export_flow_id (flow *my_flow_key, flow_id_v my_flow_id, u8 direction) {
```
mariomac, line 182: for clarity and brevity, I'd rename values as:
- `my_flow_key` --> `key`
- `my_flow_id` --> `id`

bpf/flows.c
```diff
@@ -65,89 +116,177 @@ static inline int fill_iphdr(struct iphdr *ip, void *data_end, struct flow *flow
...
+        if (flags & TCP_FIN_FLAG || flags & TCP_RST_FLAG) {
+            /* Need to evict the entry and send it via ring buffer */
+            flow_event = bpf_ringbuf_reserve(&flows, sizeof(flow_record), 0);
+            if (!flow_event) {
+                return rc;
+            }
+            export_flow_id(&flow_event->flow_key, my_flow_id, 1);
+            // flow_event->metrics.packets = my_flow_counters->packets;
+            // flow_event->metrics.bytes = my_flow_counters->bytes;
+            // flow_event->metrics.last_pkt_ts = my_flow_counters->last_pkt_ts;
```
mariomac, line 231: If we are not going to use this code, let's remove it (we already have the Git history).

mariomac (commented):
Thanks for doing this, I have few comments while i'd like to first try it locally.

One question, what happens currently if the `xflow_metric_map_ingress` and `xflow_metric_map_egress` maps reach the MAX_ENTRIES+1 size? Is the whole map evicted via ring buffer or will we loose flow information?

bpf/flow.h
praveingk: @mariomac : When max entries are reached, newer flows are sent to userspace for aggregation via ringbuffer (as in the current method).
Have more future optimizations in this area if we observe more activity in userspace, like  Eviction can be made more aggressive. Studying Real-world ocp traces would give us an idea if optimizations are needed in that directions.

Thanks for sharing the rewriteconstants, I will check it out.

praveingk: Tried this, apparently using volatile const for map size is not supported. It fails during compilation.
 error: fields must have a constant size: 'variable length array in structure' extension will never be supported
    __uint(max_entries, ingress_max_entries);

pkg/ebpf/tracer.go
```diff
@@ -187,6 +221,208 @@ func (m *FlowTracer) Unregister() error {
...
 }
 
+// Aggregates and consolidates the record values obtained the Per-CPU Hash Map
+func aggregateValues(mapValues []recordValue) recordValue {
+	var aggRecord recordValue
+	// Iterates over the array of record values and aggregates them into a single record
+	for _, mapValue := range mapValues {
+		aggRecord.Packets += mapValue.Packets
+		aggRecord.Bytes += mapValue.Bytes
+		if mapValue.FlowStartTime != 0 {
```
mariomac, line 231: Not really sure how this works, but should it be something like...?:

```go
if mapValue.FlowStartTime != 0 && mapValue.FlowStartTime < aggRecord.FlowStartTime {
```

praveingk: FlowStartTime is set only once

> I think there is something wrong in the way the packets are aggregated or flushed.
> 
> Experiment: I send an individual ICMP packet to a known IP; then I expect that it appears in the UI; then I send another packet. I'd expect that for each submission I get reported one small ongoing packet and one small returned incoming packet.
> 
> This is what I get with the old eBPF agent:
> 
> image
> 
> With this version, it seems that the packet numbers and sizes are somewhat random:
> 
> image

Thats weird, I tested with flowlogs-dump-collector, similar test, The agent is attached on a single interface.

```
$ go run flowlogs-dump-collector.go --listen_port=9999
starting flowlogs-dump-collector on port 9999
ipv4: 15:14:23.251708 ens6f0np0 IP 10.10.10.2:0 > 10.10.10.1:0: protocol:icmp dir:1 bytes:98 packets:1 ends: 15:14:23.251708
ipv4: 15:14:23.251983 ens6f0np0 IP 10.10.10.1:0 > 10.10.10.2:0: protocol:icmp dir:0 bytes:98 packets:1 ends: 15:14:23.251983
ipv4: 15:14:21.484647 ens6f0np0 IP 0.0.0.0:68 > 255.255.255.255:67: protocol:udp dir:0 bytes:333 packets:1 ends: 15:14:21.484647
arp: 15:14:28.458205 ens6f0np0 IP 0.0.0.0:0 > 0.0.0.0:0: protocol: dir:0 bytes:60 packets:1 ends: 15:14:28.458205
arp: 15:14:28.458234 ens6f0np0 IP 0.0.0.0:0 > 0.0.0.0:0: protocol: dir:1 bytes:42 packets:1 ends: 15:14:28.458234
ipv4: 15:14:39.142207 ens6f0np0 IP 10.10.10.2:0 > 10.10.10.1:0: protocol:icmp dir:1 bytes:98 packets:1 ends: 15:14:39.142207
ipv4: 15:14:39.142396 ens6f0np0 IP 10.10.10.1:0 > 10.10.10.2:0: protocol:icmp dir:0 bytes:98 packets:1 ends: 15:14:39.142396
```
The eviction from map timeout is 5 seconds, So you can expect multiple ICMP packets to be bunched if they are within that interval. However, in between the 98 byte packets, the other packets are weird. Do you see them in flowlogs-dump-collector as well?

bpf/flows.c
```diff
@@ -19,43 +47,64 @@
...
-            flow->transport.src_port = __bpf_ntohs(tcp->source);
-            flow->transport.dst_port = __bpf_ntohs(tcp->dest);
+            flow_id->src_port = __bpf_ntohs(tcp->source);
+            flow_id->dst_port = __bpf_ntohs(tcp->dest);
+            if (tcp->fin) {
+                *flags= *flags | TCP_FIN_FLAG;
+            }
+            if (tcp->rst) {
+                *flags= *flags | TCP_RST_FLAG;
+            }
```
jotak, line 100: should we add more than fin/rst? SYN maybe ?

bpf/flows.c
```diff
@@ -95,59 +151,137 @@ static inline int fill_ip6hdr(struct ipv6hdr *ip, void *data_end, struct flow *f
...
+    } else if (flow_id->eth_protocol == ETH_P_IPV6) {
         struct ipv6hdr *ip6 = (void *)eth + sizeof(*eth);
-        return fill_ip6hdr(ip6, data_end, flow);
+        return fill_ip6hdr(ip6, data_end, flow_id, flags);
+    } else {
+        // TODO : Need to implement other specific ethertypes if needed
+        // For now other parts of flow id remain zero
+        memset (&(flow_id->src_ip),0, sizeof(struct in6_addr));
+        memset (&(flow_id->dst_ip),0, sizeof(struct in6_addr));
+        flow_id->protocol = 0;
```
jotak, line 173: just to understand: this is a corner case where eth_proto is neither ipv4 nor ipv6 .. in that case we don't fill ip/proto information, however `flow_id->eth_protocol` is still present and exported in flows, right?

praveingk: Yes, exactly.

bpf/flows.c
```diff
@@ -95,59 +151,137 @@ static inline int fill_ip6hdr(struct ipv6hdr *ip, void *data_end, struct flow *f
...
+    }
 
-    struct ethhdr *eth = data;
-    if (fill_ethhdr(eth, data_end, flow) == DISCARD) {
-        bpf_ringbuf_discard(flow, 0);
+    if (my_flow_counters != NULL) {
+        my_flow_counters->packets += 1;
+        my_flow_counters->bytes += skb->len;
+        my_flow_counters->last_pkt_ts = current_time;
+        if (flags & TCP_FIN_FLAG || flags & TCP_RST_FLAG) {
```
jotak, line 220: Is it ok for long lasting connections? or should we add other eviction criteria, like what exists in ovs/ipfix, based on a packets limit reached, and/or a timeout ?

bpf/flows.c
praveingk: Ideally, I guess that information could be inferred by FlowStartTime variable. But fundamentally, yes,  We can add more flags based on the use-case.

bpf/flows.c
praveingk: Currently, I have implemented a timeout logic for flows to be evicted from userspace. Please Refer [tracer.go](https://github.com/netobserv/netobserv-ebpf-agent/pull/39/files#diff-f3248f855cc319b6d96125842a513586d6d8cbfb1e241a93e79911d08e41c728). We can potentially add other eviction logic there.

bpf/flows.c
```diff
@@ -95,59 +151,137 @@ static inline int fill_ip6hdr(struct ipv6hdr *ip, void *data_end, struct flow *f
...
+        if (direction == INGRESS) {
+            ret = bpf_map_update_elem(&xflow_metric_map_ingress, &id, &new_flow_counter,
+                                      BPF_NOEXIST);
+        } else {
+            ret = bpf_map_update_elem(&xflow_metric_map_egress, &id, &new_flow_counter,
+                                      BPF_NOEXIST);
+        }
+        if (ret < 0) {
+            /*
+                When the map is full, we have two choices:
```
jotak, line 252: There's probably an interesting discussion to have here.
I guess the two suggested options both involve a drop in performance. Moreover, when the map is full, we could assume there's high probability that it will stay full for a while (?) so it's like we're entering in a degraded mode.
What about other options:
1. Flushing the whole map when it's full, so that performance gains kinda "restart from scratch"
2. Just drop the flow?

In all cases, it would be nice to be able to report this kind of information to the ebpf agent (userspace) so that, maybe, it could emit some metrics about the state of the bpf maps? It could be useful to track performance gains and drops, to see if it correlates with maps being full ?

bpf/flows.c
```diff
@@ -95,59 +151,137 @@ static inline int fill_ip6hdr(struct ipv6hdr *ip, void *data_end, struct flow *f
...
+int ingress_flow_parse (struct __sk_buff *skb) {
+    return flow_monitor(skb, INGRESS);
 }
 
-char __license[] SEC("license") = "GPL";
+SEC("tc_egress")
+int egress_flow_parse (struct __sk_buff *skb) {
+    return flow_monitor(skb, EGRESS);
+}
+char _license[] SEC("license") = "GPL";
```
jotak, line 287: It was previously `__license` (double underscore). Does it change anything, or this is purely informative / conventional?
As a side-note, I started a discussion on slack about licensing. Not sure if we want to use GPL or another GPL-compatible one.

pkg/ebpf/tracer.go
```diff
@@ -63,6 +96,7 @@ func (m *FlowTracer) Register() error {
 		return fmt.Errorf("rewriting BPF constants definition: %w", err)
 	}
 	if err := spec.LoadAndAssign(&m.objects, nil); err != nil {
+		fmt.Printf("%s\n", m.objects)
```
jotak, line 99: debug log to remove?

mariomac:
@jotak I initially kept them there for simplification of the pipeline, but at some point we should remove them and generate them on each build.

It's important to keep them in that folder, as the generated sources assume that the .o files are in the same folders, and at the same time they are full of private functions that need to be in the same place as the client code.

bpf/flows.c
jotak: It seems to me that yes, it's necessary to have also the timeout/max-size eviction mechanism, because else data could be retained on the cpu maps potentially indefinitely, regardless the cache timeout/max-size configured. (Unless there is a mechanism that allows flushing kernel-space maps triggered from the user-space, but I didn't see that, right?)

Or maybe there's something I don't understand, I'm not yet super familiar with the agent code

pkg/ebpf/tracer.go
praveingk: Removed

bpf/flows.c
praveingk: @jotak : SEC("license") is the declaration of license, but the variable "char license[]" can be anything. Generally, I noticed "_license" being used (ref :https://github.com/cilium/tetragon/blob/main/bpf/process/bpf_generic_kprobe.c#L17)

praveingk:
> I think there is something wrong in the way the packets are aggregated or flushed.
> 
> Experiment: I send an individual ICMP packet to a known IP; then I expect that it appears in the UI; then I send another packet. I'd expect that for each submission I get reported one small ongoing packet and one small returned incoming packet.
> 
> This is what I get with the old eBPF agent:
> 
> image
> 
> With this version, it seems that the packet numbers and sizes are somewhat random:
> 
> image

Thanks @mariomac for checking this. Upon verifying the setup on kubernetes environment, I found that hash-collisions were leading to multiple flows mapping on to the same map entry, leading to inaccurate aggregation. I will be working on resolving this, by keeping track of flowid in the entries, and modifying the eviction logic and aggregation logic.

praveingk:
> 

@mariomac : I have fixed the issue of hash collision. Verfied on an openshift cluster and e2e-tests. Please refer to the commits to README to know more about the fix.

pkg/ebpf/tracer.go
```diff
@@ -7,16 +7,29 @@ import (
...
+#include <time.h>
+static unsigned long long get_nsecs(void)
+{
+    struct timespec ts;
+    clock_gettime(CLOCK_MONOTONIC, &ts);
+    return (unsigned long long)ts.tv_sec * 1000000000UL + ts.tv_nsec;
+}
+*/
+import "C"
+
```
mariomac, line 32: That would force us to enable CGO (https://dave.cheney.net/2016/01/18/cgo-is-not-go).

Can't we just remove this and replace:
```go
timeNow := flow.Timestamp(C.get_nsecs())
```
by:
```go
timeNow := time.Now().UnixNano()
```
?

praveingk: The problem with the timestamp is that eBPF programs use CLOCK_MONOTONIC timestamp, however, the timeGo uses CLOCK_REALTIME. Hence I used the C wrapper.

`DEBU[0003] Monitor                                       component=ebpf.FlowTracer iface=ens6f0np0 timeNow=524694462887368 timeNowGo=1659419741836160042`

Above code snippet returns both time using C.get_nsecs() and time.Now().UnixNano(). Let me check if there is monotonic clock available in go.

mariomac: According to https://pkg.go.dev/time

> The Time returned by time.Now contains a monotonic clock reading.

mariomac: Ah, wait a minute, it's true that time.Now().UnixNano() is not monotonic.

The way of getting monotonic reads is by substracting two timestamps. e.g. subtract time.Now() to the agent start time:
https://pkg.go.dev/time#hdr-Monotonic_Clocks

```
start := time.Now()
... operation that takes 20 milliseconds ...
t := time.Now()
elapsed := t.Sub(start)
```

praveingk: As we had discussed, I have used monotime module. I used the package https://pkg.go.dev/github.com/gavv/monotime

bpf/flows.c
praveingk: We will definitely need to discuss more about which flags could be captured.
it could be added as additional enhancements on top?

pkg/ebpf/tracer.go
```diff
@@ -16,20 +16,10 @@ import (
 	"github.com/netobserv/netobserv-ebpf-agent/pkg/flow"
 	"github.com/sirupsen/logrus"
 	"github.com/vishvananda/netlink"
+	"github.com/gavv/monotime"
```
mariomac: Maybe running "go mod vendor" and pushing the new dependencies will fix this compilation error.

praveingk: Done.

pkg/ebpf/tracer.go
```diff
@@ -187,9 +220,310 @@ func (m *FlowTracer) Unregister() error {
...
+				}
+				mapEntries++
+				aggRecord, collidedRecords := m.aggregateValues(mapKey, mapValues)
+
+				// Eviction Logic can be based on the following three metrics:
+				//   1) LastPacket > n sec ?
+				//   2) Packet/Byte count > Threshold ?
+				//   3) Number of Map entries > Threshold
+				// 			Perform more aggressive eviction by reducing the flow timeout duration
+				//   4) Regular Eviction (every n seconds)
```
mariomac, line 440: I have been digging into the code and maybe we don't need all the cases here, so the code would be simple and more stable.
E.g.:
- Can we delete case 1, considering that, in case 4 we will evict all flows every n seconds anyway.
- Can we delete case 2? I guess there is nothing wrong with a flow with many packets/bytes count.

So I'd evict as for the previous version of the agent:
1) When the number of map entries > CacheMaxFlows
2) Regularly, every CacheActiveTimeout

pkg/agent/config.go
```diff
@@ -69,4 +69,6 @@ type Config struct {
 	KafkaCompression string `env:"KAFKA_COMPRESSION" envDefault:"none"`
 	// ProfilePort sets the listening port for Go's Pprof tool. If it is not set, profile is disabled
 	ProfilePort int `env:"PROFILE_PORT"`
+	// EvictionTimeout sets the timeout for eviction of an inactive flow, Default is 2 seconds
+	EvictionTimeout uint64 `env:"EVICTION_TIMEOUT" envDefault:"2000000000"`
```
mariomac, line 73: Configuration values in nanoseconds are very unfriendly to the users. I did this PR against your own PR to use durations:

https://github.com/netobserv/netobserv-ebpf-agent/pull/45/files

mariomac:
It seems that e2e tests still might fail from time to time. Investigating more about this: https://github.com/netobserv/netobserv-ebpf-agent/runs/7645373651?check_suite_focus=true

bpf/flows.c
```diff
@@ -95,59 +151,186 @@ static inline int fill_ip6hdr(struct ipv6hdr *ip, void *data_end, struct flow *f
...
+            flow_event->metrics.bytes = skb->len;
+            flow_event->metrics.last_pkt_ts = current_time;
+            bpf_ringbuf_submit(flow_event, 0);
+            // Defer the deletion of the entry from the map to usespace since it evicts other CPU metrics
+        } else {
+            bool update = false;
+            my_flow_aggregate->packets += 1;
+            my_flow_aggregate->bytes += skb->len;
+            my_flow_aggregate->last_pkt_ts = current_time;
+            if (my_flow_aggregate->eth_protocol == 0) { // Ensures no other flow is residing
```
mariomac, line 228: Under which circumstance it could happen that `my_flow_aggregate != NULL` (it has been stored in the map) but the `eth_protocol` is 0 (as if no flow was stored)?

bpf/flows.c
praveingk: I am currently using FIN/RST flag to communicate to the userspace agent that, the flow is complete. This will trigger the userspace agent to evict the entries of the flows immediately, instead of waiting for the timeout. 
This is done mainly to evict entries immediately instead of waiting until evictionTimeout to clear the map space for other flows.

bpf/flows.c
praveingk: This happens when a flow X from CPU 1 has already created the key and entry. Now, When flow Y from CPU2 hashes into the same bucket due to hash collision, this condition happens (i.e. flow_aggregate is not null)
In these cases, we still allow the flow Y to reside in the entry, and handle the collidedRecords in userspace.

Ofcourse, when flow X and flow Y are from the same CPU, we cannot let flow Y update the same bucket where flow X has active entries.

bpf/flows.c
```diff
@@ -95,59 +151,186 @@ static inline int fill_ip6hdr(struct ipv6hdr *ip, void *data_end, struct flow *f
...
+                my_flow_aggregate->dst_port = id.dst_port;
+                my_flow_aggregate->protocol = id.protocol;
+                update = true;
+            } else {
+                // check flow id stored
+                // If its the same, then perform write to map
+                // Else send to ringbuffer
+                if (my_flow_aggregate->src_port == id.src_port &&
+                    my_flow_aggregate->dst_port == id.dst_port &&
+                    my_flow_aggregate->protocol == id.protocol) {
```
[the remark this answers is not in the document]
praveingk: I think my previous comment explains this question.

bpf/flows.c
praveingk: Yes, currently, I constantly monitor the size of the maps, no. of collisions etc. Which in future, we can use to take wise decisions.

bpf/flows.c
```diff
@@ -95,59 +151,186 @@ static inline int fill_ip6hdr(struct ipv6hdr *ip, void *data_end, struct flow *f
...
+            /*
+                When the map is full, we have two choices:
+                    1) Send the new flow entry to userspace via ringbuffer,
+                       until an entry is available.
+                    2) Send an existing flow entry (probably least recently used)
+                       to userspace via ringbuffer, delete that entry, and add in the
+                       new flow to the hash map.
+
+                Ofcourse, 2nd step involves more manipulations and
+                       state maintenance, and no guarantee if it will any performance benefit.
```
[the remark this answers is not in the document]
praveingk: Agree. I will remove the option 2 from the comment, and maybe move it to design choices in readme.

pkg/agent/config.go
praveingk: Thanks for adding this.

bpf/flows.c
mariomac: As long as our product is not (yet) connection aware, I think this is not needed as it won't provide any extra information/input to the users. We actually follow the "standard" definition of flow that do not consider the connection status: https://datatracker.ietf.org/doc/html/rfc3954#section-2

> a set of IP packets passing an Observation Point in the network during a certain time interval.

The code it's correct but feel free to remove it if you think it can clean/simplify the code (I'd do it)

bpf/flows.c
mariomac: I see. I must have misunderstood the meaning of `BPF_MAP_TYPE_PERCPU_HASH`. I thought it created a different map for each CPU so no possibilities of colliding two entries from the same CPU.

pkg/ebpf/tracer.go
praveingk: Agree. I have just enumerated the options that could be implemented in future.

praveingk:
@mariomac : As suggested, I have removed the TCP Flags from the bpf code to stick to flow-level definition.

mariomac (approved):

PR #41, opened 2022-07-04
add IPAddrFromNetIP function to allow convertion from NetIP to IPAddr

eranra:


pkg/flow/record.go
```diff
@@ -73,6 +73,13 @@ func (r *Record) Accumulate(src *Record) {
 	r.Packets += src.Packets
 }
 
+// IPAddrFromNetIP returns IPAddr from net.IP
+func IPAddrFromNetIP(netIP net.IP) IPAddr {
```
mariomac, line 77: since this is a function that is only used in a test, shall we move it to the `kafja_json_test.go` file?

mariomac (approved):

PR #43, opened 2022-07-11
Increased number of ICMP test packets

mariomac:
In order to ensure that the tests will fail if ICMP packets aren't accounted correctly, it increases the number of tested packets.


praveingk (approved):
lgtm

PR #51, opened 2022-09-02
NETOBSERV-560: fix fragmentation of reported flows

mariomac:
The unsynchronized access to eBPF maps might cause that the flows' hashmap contains multiple entries for the same key, so big flows (e.g. downloading a large file) were submitted as multiple, small-sized flows.

We support Kernel 4.18 and synchronization primitives are not introduced until Kernel 5.1.

The solution has been to aggregate at userspace the flows that share the same identifier. No noticeable impact in performance nor memory usage.


jpinsonneau (approved):
Tested and LGTM. I only wonder what should be the default cache parameter value ?

I see `cacheMaxFlows: 1000` in the sample config of the operator:
https://github.com/netobserv/network-observability-operator/blob/main/config/samples/flows_v1alpha1_flowcollector.yaml#L17
whereas it's `env:"CACHE_MAX_FLOWS" envDefault:"5000"` in ebpf-agent config:
https://github.com/netobserv/netobserv-ebpf-agent/blob/main/pkg/agent/config.go#L35

mariomac:
@jpinsonneau Good question. I raised up to 5000 because I observed that in a small OpenShift cluster (3 masters + 3 workers), the flows are always evicted sooner because the 1000-flows map is filled before the 5 seconds of eviction period

PR #52, opened 2022-09-07
Update measurements with minor details

praveingk:
Added details on the measurement tools used, packets size, etc.


mariomac (approved):
Nice numbers! thanks

PR #53, opened 2022-09-08
Add TCP traffic measurements

praveingk:
Included TCP measurements


mariomac (approved):

praveingk:
Basically, I send 100B packets (by controlling mtu of the interface) and run 128 parallel iperf3s. The traffic runs for 60 sec, and I monitor the throughput, overall packets sent, packets monitered by the agent, cpu ,etc:

- Without any agent, the overall packet rate reaches ~11 Mpps
- With V1 agent, the packet rate drops to 3.7 Mpps, however with V2 it reaches 10.4 Mpps
- With pmacct its about 3.5 Mpps, and tcpdump is about 7.25 Mpps
- The monitoring rate for V2 agent is 100%, but it's much lower for others
- The additional CPU overhead for V2 agent incurred is about 4%, but is much higher for others

PR #55, opened 2022-09-23
NETOBSERV-589: send kafka flows as protobuf lines

mariomac:
This PR needs to be merged together with: https://github.com/netobserv/network-observability-operator/pull/171

Breaking changes:

* For Kafka emitter, replaces the JSON encoding by Protobuf encoding. The reason to not keep the old one is that we won't need JSON as soon as Flowlogs-Pipeline accepts Protobuf. We first need this PR to be merged: https://github.com/netobserv/flowlogs-pipeline/pull/313
* The `KAFKA_BATCH_BYTES` is removed and replaced by the `KAFKA_BATCH_SIZE` variable. The previous `KAFKA_BATCH_SIZE` variable has been now replaced by the development-only `KAFKA_BATCH_MESSAGES` variable. 	
 The reason is that the "standard" Kafka name for the variable that sets the payload size in bytes is "batch.size". We have
 decided to hide the Segmentio's BatchSize meaning (payload length in messages), which at the end is an implementation detail, and expose to the users the common, standard meaning for batch.size.


pkg/agent/config.go
```diff
@@ -55,12 +55,13 @@ type Config struct {
 	KafkaBrokers []string `env:"KAFKA_BROKERS" envSeparator:","`
 	// KafkaTopic is the name of the topic where the flows' processor will receive the flows from.
 	KafkaTopic string `env:"KAFKA_TOPIC" envDefault:"network-flows"`
-	// KafkaBatchSize sets the limit on how many messages will be buffered before being sent to a
+	// KafkaBatchMessages sets the limit on how many messages will be buffered before being sent to a
 	// partition.
-	KafkaBatchSize int `env:"KAFKA_BATCH_SIZE" envDefault:"100"`
-	// KafkaBatchBytes sets the limit of the maximum size of a request in bytes before being sent
+	KafkaBatchMessages int `env:"KAFKA_BATCH_MESSAGES" envDefault:"1000"`
```
jpinsonneau, line 60: do we get better results with `1000` here ?

pkg/agent/config.go
```diff
@@ -55,12 +55,13 @@ type Config struct {
...
+	// KafkaBatchSize sets the limit, in bytes, of the maximum size of a request before being sent
 	// to a partition.
-	KafkaBatchBytes int `env:"KAFKA_BATCH_BYTES" envDefault:"1048576"`
+	KafkaBatchSize int `env:"KAFKA_BATCH_SIZE" envDefault:"1048576"`
+	// KafkaLingerMS sets the amount of time to wait before sending a
```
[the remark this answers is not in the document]
mariomac: Oh yes, I forgot to remove the comment. I initially tried to configure linger but had to remove the argument when I realized that the Go Kafka library does not allow to set it.

pkg/agent/config.go
mariomac: Yes, a small, idle cluster is generating about 500 flows on each period. That would mean 5 batches to be sent to Kafka. For big clusters, we observed around 10.000 flows on each batch (around 100 batches to Kafka). Increasing that to 1000 would mean less interactions with kafka (1 and 10 batches, respectively)

jpinsonneau: makes sense :+1:

jpinsonneau (approved):
LGTM on code side. Thanks @mariomac !

jotak (approved):

PR #56, opened 2022-09-29
NETOBSERV-578: deduplicate flows at agent level

mariomac:
Detects when the same flow is detected in different interfaces and then forwards the flow from the first interface it was received from.

In hosts with low traffic it decreased the number of flows by 20% and hosts with higher traffic, by 40%.

However, it seems that in OpenShift we still could have some duplicates as the same flow can be detected from two agents in different hosts. In that case, a second deduplication should be performed at flowlogs-pipeline level.

### DIFF
--- a/pkg/agent/config.go
+++ b/pkg/agent/config.go
@@ -47,7 +47,8 @@ type Config struct {
 	// a flow hasn't been received for that expiry time, the deduplicator forgets it. That means
 	// that a flow from a connection that has been inactive during that period could be forwarded
 	// again from a different interface.
-	DeduperFCExpiry time.Duration `env:"DEDUPER_FC_EXPIRY" envDefault:"30s"`
+	// If the value is not set, it will default to 2 * CacheActiveTimeout
+	DeduperFCExpiry time.Duration `env:"DEDUPER_FC_EXPIRY"`
 	// Logger level. From more to less verbose: trace, debug, info, warn, error, fatal, panic.
 	LogLevel string `env:"LOG_LEVEL" envDefault:"info"`
 	// Sampling holds the rate at which packets should be sampled and sent to the target collector.
--- a/pkg/agent/config.go
+++ b/pkg/agent/config.go
@@ -5,8 +5,10 @@ import (
 )
 
 const (
-	ListenPoll  = "poll"
-	ListenWatch = "watch"
+	ListenPoll       = "poll"
+	ListenWatch      = "watch"
+	DeduperNone      = "none"
+	DeduperFirstCome = "firstCome"
 )
 
 type Config struct {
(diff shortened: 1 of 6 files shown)


pkg/flow/deduper.go
```diff
@@ -0,0 +1,100 @@
...
+// deduperCache implement a LRU cache whose elements are evicted if they haven't been accessed
+// during the expire duration.
+// It is not safe for concurrent access.
+type deduperCache struct {
+	expire time.Duration
+	// key: RecordKey with the interface and MACs erased, to detect duplicates
+	// value: listElement pointing to a struct entry
+	ifaces map[RecordKey]*list.Element
+	// element: entry structs of the ifaces map ordered by expiry time
+	ll *list.List
```
OlivierCazade, line 22: Nitpicking: may be ll is too generic, may be something like RecRefs?

pkg/flow/deduper.go
```diff
@@ -0,0 +1,100 @@
...
+		ifaces: map[RecordKey]*list.Element{},
+	}
+	return func(in <-chan []*Record, out chan<- []*Record) {
+		for records := range in {
+			cache.removeExpired()
+			fwd := make([]*Record, 0, len(records))
+			for _, record := range records {
+				if !cache.isDupe(&record.RecordKey) {
+					fwd = append(fwd, record)
+				}
```
OlivierCazade, line 47: Possible improvement: adding a metric here to count the number of duplicate.

pkg/agent/config.go
```diff
@@ -36,6 +38,16 @@ type Config struct {
...
+	// Deduper specifies the deduper type. Accepted values are "none" (disabled) and "firstCome".
+	// When enabled, it will detect duplicate flows (flows that have been detected e.g. through
+	// both the physical and a virtual interface).
+	// "firstCome" will forward only flows from the first interface the flows are received from.
+	Deduper string `env:"DEDUPER" envDefault:"none"`
+	// DeduperFCExpiry specifies the expiry duration of the flows "firstCome" deduplicator. After
+	// a flow hasn't been received for that expiry time, the deduplicator forgets it. That means
+	// that a flow from a connection that has been inactive during that period could be forwarded
+	// again from a different interface.
+	DeduperFCExpiry time.Duration `env:"DEDUPER_FC_EXPIRY" envDefault:"30s"`
```
OlivierCazade, line 50: Isn't the 30s default a bit too high causing unnecessary memory consumption?

For a same flow being captured going through 2 different interface on the same node, I would have thought that this was below the second. So may be a default value like 5s would be enough?

OlivierCazade (approved):
LGTM, just a few comments but they are not blocking comments feel free to ignore.

pkg/agent/config.go
mariomac: Hmm since I want to be consistent between flows' flush periods, this should be > 1 cache flush periods. So probably we can set it, by default, to "CacheActiveTimeout * 2" (10s)

pkg/flow/deduper.go
mariomac: I added a debug log below. If you mean prometheus metrics, maybe we can create a new task to integrate a metrics server here.

OlivierCazade (approved):

PR #63, opened 2022-10-13
NETOBSERV-613: drop messages when they accumulate in the exporter

mariomac:
The following PRs need to be merged first: https://github.com/netobserv/netobserv-ebpf-agent/pull/61 and https://github.com/netobserv/netobserv-ebpf-agent/pull/62

This PR adds a flow's dropper before the exporter to avoid memory growing indefinitely or synchronized channel operations blocking previous parts of the system.

It also logs the number of dropped flows, suggesting the customer to take reactive actions.

The new architecture results in:

```mermaid
flowchart TD

    E(ebpf.FlowFetcher) --> |"pushes via<br/>RingBuffer"| RB(flow.RingBufTracer)
    E <--> |"polls<br/>PerCPUHashMap"| M(flow.MapTracer)
    RB --> |chan *flow.Record| ACC(flow.Accounter)

    ACC --> |"chan []*flow.Record"| DD(flow.Deduper)
    M --> |"chan []*flow.Record"| DD

    subgraph Optional
        DD
    end

    DD --> |"chan []*flow.Record"| CL(flow.CapacityLimiter)

    CL --> |"chan []*flow.Record"| EX("export.GRPCProto<br/>or<br/>export.KafkaProto")

```

### DIFF
--- a/pkg/flow/limiter.go
+++ b/pkg/flow/limiter.go
@@ -1,0 +1,59 @@
+package flow
+
+import (
+	"time"
+
+	"github.com/sirupsen/logrus"
+)
+
+const initialLogPeriod = time.Minute
+const maxLogPeriod = time.Hour
+
+var cllog = logrus.WithField("component", "capacity.Limiter")
+
+// CapacityLimiter forwards the flows between two nodes but checks the status of the destination
+// node's buffered channel. If it is already full, it drops the incoming flow and periodically will
+// log a message about the number of lost flows.
+type CapacityLimiter struct {
+	droppedFlows int
+}
+
+func (c *CapacityLimiter) Limit(in <-chan []*Record, out chan<- []*Record) {
+	go c.logDroppedFlows()
+	for i := range in {
+		if len(out) < cap(out) {
+			out <- i
+		} else {
+			c.droppedFlows += len(i)
+		}
+	}
+}
+
+func (c *CapacityLimiter) logDroppedFlows() {
+	logPeriod := initialLogPeriod
+	debugging := logrus.IsLevelEnabled(logrus.DebugLevel)
+	for {
+		time.Sleep(logPeriod)
+
+		// a race condition might happen in this counter but it's not important as it's just for
+		// logging purposes
+		df := c.droppedFlows
+		if df > 0 {
+			c.droppedFlows = 0
+			cllog.Warnf("%d flows were dropped during the last %s because the agent is forwarding "+
+				"more flows than the remote ingestor is able to process. You might "+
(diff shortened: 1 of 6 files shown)


pkg/flow/limiter.go
```diff
@@ -0,0 +1,59 @@
...
+	debugging := logrus.IsLevelEnabled(logrus.DebugLevel)
+	for {
+		time.Sleep(logPeriod)
+
+		// a race condition might happen in this counter but it's not important as it's just for
+		// logging purposes
+		df := c.droppedFlows
+		if df > 0 {
+			c.droppedFlows = 0
+			cllog.Warnf("%d flows were dropped during the last %s because the agent is forwarding "+
```
jotak, line 43: maybe it's time to have metrics rather than logs, for this kind of things? Or create a follow-up ?

pkg/flow/limiter.go
```diff
@@ -0,0 +1,59 @@
...
+var cllog = logrus.WithField("component", "capacity.Limiter")
+
+// CapacityLimiter forwards the flows between two nodes but checks the status of the destination
+// node's buffered channel. If it is already full, it drops the incoming flow and periodically will
+// log a message about the number of lost flows.
+type CapacityLimiter struct {
+	droppedFlows int
+}
+
+func (c *CapacityLimiter) Limit(in <-chan []*Record, out chan<- []*Record) {
```
jotak, line 21: I'm not sure it's worth it to have a whole stage with in/out buffers for just this simple check, no ? I tried benchmarking against another version with the limiter done at the source and got slightly better time: 15016 ns/op versus 21698 ns/op with a limiter stage (no change in number of allocs)

pkg/flow/limiter.go
mariomac: I agree with you despite we can also keep this message as a first-hand notification that could be useful for the user. I'd create a JIRA issue to implement and expose metrics.

pkg/flow/limiter.go
mariomac: Yeah, I just wanted to separate each functionality as much as possible to avoid overloading too much some pipeline stages. Also because otherwise we should implement this in multiple places: since the deduper is optional, it should Go both in the FlowAccounter and MapTracer.

About the performance improvement, I think in this case is negligible as the pipeline processes in batches and each batch is usually forwarded every `CacheActiveTimeout` period (`5s` by default).

jotak: fair enough, after our discussion I understand better.
Anyway we can still revisit later if we find there's optimizations to bring here

jotak (approved):

Issue #26, opened 2022-05-11
Add support for export using IPFIX / Netflow


eranra:
This will remove the strong bond between the agent and FLP which is currently using a non-standard and proprietary gRPC protocol and allows the agent the be used by other tools that work with standard IPFIX / Netflow. For an upstream project, this is needed if we are looking for external contributions.

eranra:
ref: https://github.com/netobserv/netobserv-ebpf-agent/blob/main/pkg/exporter/grpc_proto.go

praveingk:
This is now addressed in https://github.com/netobserv/netobserv-ebpf-agent/pull/64

PR #64, opened 2022-10-20
Enable IPFIX export

praveingk:
This PR address the below issue [https://github.com/netobserv/netobserv-ebpf-agent/issues/26](url) to provide IPFIX export of flow records, which can be consumed directly by any platform.


mariomac:
Doing an initial view, it looks good! I'll do later a more detailed review.

In the meantime, I think we should add e2e tests. For example, you can just copy the [kafka folder](https://github.com/netobserv/netobserv-ebpf-agent/tree/main/e2e/kafka) into e.g. `ipfix` and add some modifications:

1- in the `manifests/` subfolders, remove the kafka manifests, then configure the flp-transformer and agent manifests to export/accept IPFIX data (FLP already accepts IPFIX)

2- rename the kafka_test.go to ipfix_test.go, remove [this lines](https://github.com/netobserv/netobserv-ebpf-agent/blob/main/e2e/kafka/kafka_test.go#L46-L70) and [this lines](https://github.com/netobserv/netobserv-ebpf-agent/blob/main/e2e/kafka/kafka_test.go#L96-L131).

praveingk:
> Doing an initial view, it looks good! I'll do later a more detailed review.
> 
> In the meantime, I think we should add e2e tests. For example, you can just copy the [kafka folder](https://github.com/netobserv/netobserv-ebpf-agent/tree/main/e2e/kafka) into e.g. `ipfix` and add some modifications:
> 
> 1- in the `manifests/` subfolders, remove the kafka manifests, then configure the flp-transformer and agent manifests to export/accept IPFIX data (FLP already accepts IPFIX)
> 
> 2- rename the kafka_test.go to ipfix_test.go, remove [this lines](https://github.com/netobserv/netobserv-ebpf-agent/blob/main/e2e/kafka/kafka_test.go#L46-L70) and [this lines](https://github.com/netobserv/netobserv-ebpf-agent/blob/main/e2e/kafka/kafka_test.go#L96-L131).

Thanks Mario, I will add the e2e tests.

mariomac:
Adding hold labels for the feature freeze.

praveingk:
> 
@mariomac 
I have added e2e tests. Do note that in FLP config, I had to fix "Interface" to  a non-zero field, since flp's ipfix input from goflow does not return interface name which was exported by ebpf agent.

pkg/exporter/ipfix_proto.go
```diff
@@ -0,0 +1,498 @@
...
+	"github.com/netobserv/netobserv-ebpf-agent/pkg/flow"
+	"github.com/sirupsen/logrus"
+	"github.com/vmware/go-ipfix/pkg/entities"
+	ipfixExporter "github.com/vmware/go-ipfix/pkg/exporter"
+	"github.com/vmware/go-ipfix/pkg/registry"
+)
+
+var ilog = logrus.WithField("component", "exporter/IPFIXProto")
+
+type IPFIXProto struct {
```
mariomac, line 17: I'd name this struct as IPFIX (and the file as `ipfix.go` to avoid confusion).

The `GRPCProto` name was derived from `GRPC+Protocol Buffers`. Since this doesn't use protocol buffers, you can remove the `Proto` prefix.

pkg/exporter/ipfix_proto.go
```diff
@@ -0,0 +1,498 @@
...
+func sendDataRecordv6(log *logrus.Entry, record *flow.Record, exporter *ipfixExporter.ExportingProcess, templateID uint16) error {
+	dataSet := entities.NewSet(false)
+	err := dataSet.PrepareSet(entities.Data, templateID)
+	if err != nil {
+		log.Errorf("Failed in PrepareSet")
+		return err
+	}
+	elements := make([]entities.InfoElementWithValue, 0)
+
+	err = addElementToTemplate(log, "ethernetType", makeByteFromUint16(record.EthProtocol), &elements)
```
mariomac, line 90: Is the IPFIX exporter performing correctly? It looks to me that invoking `addElementToTemplate` for each record, for each field, is something that can be optimized.

Basically because `addElementToTemplate` invokes `registry.GetInfoElement` (a lookup) and a decode... I wonder whether this can be cached somewhere.

mariomac (commented):
Mostly looks good. I have a suggestion about renaming a struct and a file, and a concern about the performance.

pkg/exporter/ipfix_proto.go
```diff
@@ -0,0 +1,498 @@
...
+	if err != nil {
+		log.WithError(err).Error("Failed in Add Record")
+		return err
+	}
+	_, err = exporter.SendSet(dataSet)
+	if err != nil {
+		log.WithError(err).Error("Failed in Send Record")
+		return err
+	}
+	log.Printf("Sending IPFIX with %+v -> %+v", record.Network.SrcAddr.IP(), record.Network.DstAddr.IP())
```
mariomac, line 164: This will flood the logs. Please set it at trace level.

pkg/exporter/ipfix_proto.go
```diff
@@ -0,0 +1,498 @@
...
+	if err != nil {
+		return err
+	}
+	err = addElementToTemplate(log, "interfaceName", []byte(record.Interface), &elements)
+	if err != nil {
+		return err
+	}
+	err = dataSet.AddRecord(elements, templateID)
+	if err != nil {
+		log.WithError(err).Error("Failed in Add Record")
```
mariomac, line 156: No need to log errors if we return the error (the function that receives it will log it again later)

pkg/exporter/ipfix_proto.go
praveingk: Sure, Will change it.

pkg/exporter/ipfix_proto.go
praveingk: Yes, you are right, I will store the set in the global struct, so that we do not need to create it again for each record.

pkg/exporter/ipfix_proto.go
```diff
@@ -0,0 +1,498 @@
...
+	err = addElementToTemplate(log, "packetDeltaCount", nil, &elements)
+	if err != nil {
+		return 0, err
+	}
+	err = addElementToTemplate(log, "interfaceName", nil, &elements)
+	if err != nil {
+		return 0, err
+	}
+
+	fmt.Printf("%+v", elements)
```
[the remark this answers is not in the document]
praveingk: Done

pkg/exporter/ipfix_proto.go
```diff
@@ -0,0 +1,498 @@
...
+// TimeFlowStart -> flowStartSeconds, flowStartNanoseconds
+// TimeFlowEnd -> flowEndSeconds, flowEndNanoseconds
+// Packets     -> packetDeltaCount
+// Interface   -> interfaceName
+
+func sendDataRecordv6(log *logrus.Entry, record *flow.Record, exporter *ipfixExporter.ExportingProcess, templateID uint16) error {
+	dataSet := entities.NewSet(false)
+	err := dataSet.PrepareSet(entities.Data, templateID)
+	if err != nil {
+		log.Errorf("Failed in PrepareSet")
```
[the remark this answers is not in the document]
praveingk: Done

pkg/exporter/ipfix_proto.go
```diff
@@ -0,0 +1,498 @@
...
+		return err
+	}
+	err = dataSet.AddRecord(elements, templateID)
+	if err != nil {
+		log.WithError(err).Error("Failed in Add Record")
+		return err
+	}
+	_, err = exporter.SendSet(dataSet)
+	if err != nil {
+		log.WithError(err).Error("Failed in Send Record")
```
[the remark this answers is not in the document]
praveingk: Done

pkg/exporter/ipfix_proto.go
praveingk: I have cached the ieValues now.

mariomac (approved):

PR #66, opened 2022-10-26
NETOBSERV-619: mark duplicate flows

mariomac:
When the Deduper is configured and the `DEDUPER_JUSTMARK` env var is set to true, it just detects duplicate flows and marks them with a new `Duplicate` boolean values, leaving only one of the duplicates as `Duplicate: false`.

It also adds deduplication to flows with same source/destination date but different direction. This means that flows between two pods in the same node will be now deduplicated, but flows between pods in different nodes will be still reported as duplicate (as there is no way to deduplicate flows that have been captured by different agents). Before this change, same ingress/egress flow in the same node was duplicate, and flows between two nodes were quadruplicate. 

This PR does not complete the NETOBSERV-619 task. Still missing:
- Encode the duplicate field in the FLP side.
- Change the NOO to configure by default the deduper with the justmark flag (allow optional disable).
- Change the console to ignore the duplicate flows in the rate calculations.


mariomac:
@jotak yeah, you are right. Commit [df196ca](https://github.com/netobserv/netobserv-ebpf-agent/pull/66/commits/df196ca9f72d7e84f242d0c9173b20cb64c10fd8) reverts back the Direction deduplication.

OlivierCazade (approved):
LGTM, not tested, but will be tested once we implement other parts of the story.

OlivierCazade:
@mariomac , also, my understanding here, is that this PR address duplicate accross a single node, but does not address duplicate of flow reported by multiple agents, is that right?

mariomac:
@OlivierCazade exactly. Actually the whole NETOBSERV-619 is about deduplicating flows per node, not on a global basis.

In a flow going from node A to node B, it will be captured as egress flow in A and ingress flow in B. However, this is taken into account for the metrics calculation so it's fine to have it like that, at the moment.

PR #68, opened 2022-11-21
Added TCP Flags to Flow Record metrics

shach33:
I added a 16 bit flag to every record. The TCP header is read and 8 flags (fin, syn, rst, psh, ack, urg, ece, cwr) are added to bits 0-7 in flags member of metrics struct. 
During accumulation, the flags member is bitwise OR'ed to get all flags in bits 0-7.


mariomac:
@slach33 Thank you for your contribution! In order to facilitate the C code review, could you keep the original indent configuration? It's 4 spaces, no tab.

eranra:
@shach33
 Thanks for pushing the PR :slightly_smiling_face: ---- I wanted to ask about the indent changes --- seems to me that there are multiple changes that are due to indent configuration and this makes it hard to review

@mariomac @praveingk ^^^  any specific common way to fix that ???

proto/flow.proto
```diff
@@ -32,7 +32,8 @@ message Record {
 
   uint64 bytes = 8;
   uint64 packets = 9;
-  string interface = 10;
+  uint32 flags = 10;
+  string interface = 11;
```
mariomac, line 36: Changing the interface index would create breaking changes in flowlogs-pipeline.

This should be:
```
string interface = 10;
uint32 flags = 11;
```

ronensc:
@shach33 Thank you very much for this PR! :)
I am interested in the TCP flags to allow the connection tracking module in FLP to identify which of the endpoints started the connection (i.e. who sent the first SYN).

IIUC accumulating all the flags of the flowlogs with bitwise OR won't allow me to find which endpoint started the connection.

As known, in TCP the 3-step handshake goes like this:
1. client -> server: SYN
2. server -> client: SYN+ACK
3. client -> server: ACK

I think that it is most likely that the packets of step 1 and 3 will end up in the same flowlog. In this case, that flowlog will have both SYN and ACK flags making it indistinguishable from the flowlog of step 2.
Please correct me if I'm wrong.

cc @eranra @jotak @mariomac

ronensc:
Thanks @praveingk
I had a discussion with @eranra about your suggestion.
We think that I might work for the use case of identifying the first SYN. But, if we set the SYN flag in a flowlog only when the ACK flag is not set in the packet, then the flowlog of the SYN+ACK packet won't have the SYN flag set. So using the term "TCP flags" in the context of flowlogs feels inaccurate. We can use a different term for these flags such as "start" to avoid confusion.

We also talked about that in latency measurements, we need to spot both the SYN and SYN+ACK flowlogs. And in the suggested approach this is not possible.

We suggest other alternatives:
1. Aggregate packets into flowlogs based on their TCP flags. In other words, packets with different TCP flags will end up in different flowlogs.

2. Instead of having a bitwise-or aggregation of the TCP flags, having a list of the unique TCP flags combination. For example, if a flowlog consists of 3 packets with the flags `[SYN, ACK, ACK]` then the flowlog will have the 2-item list `[SYN, ACK]`. And if a flowlog consists of 3 pakcets with the flags `[SYN+ACK, ACK, ACK]` then the flowlog will have the 2-item list `[SYN+ACK, ACK]`.

praveingk:
> Thanks @praveingk I had a discussion with @eranra about your suggestion. We think that I might work for the use case of identifying the first SYN. But, if we set the SYN flag in a flowlog only when the ACK flag is not set in the packet, then the flowlog of the SYN+ACK packet won't have the SYN flag set. So using the term "TCP flags" in the context of flowlogs feels inaccurate. We can use a different term for these flags such as "start" to avoid confusion.
> 
> We also talked about that in latency measurements, we need to spot both the SYN and SYN+ACK flowlogs. And in the suggested approach this is not possible.
> 
> We suggest other alternatives:
> 
> 1. Aggregate packets into flowlogs based on their TCP flags. In other words, packets with different TCP flags will end up in different flowlogs.
> 2. Instead of having a bitwise-or aggregation of the TCP flags, having a list of the unique TCP flags combination. For example, if a flowlog consists of 3 packets with the flags `[SYN, ACK, ACK]` then the flowlog will have the 2-item list `[SYN, ACK]`. And if a flowlog consists of 3 pakcets with the flags `[SYN+ACK, ACK, ACK]` then the flowlog will have the 2-item list `[SYN+ACK, ACK]`.

@ronensc  : Yes, I understand the point, and I agree. Can you confirm if the list of combinations that can be incorporated in the eBPF code are below: 
1) SYN
2) FIN
3) RST
4) SYN+ACK
5) FIN+ACK
6) RST+ACK
7) URG
8) PSH
9) ECE
10) CWR

ronensc:
I think this list is sufficient.
@eranra WDYT?

mariomac:
@ronensc with respect to:

> Instead of having a bitwise-or aggregation of the TCP flags, having a list of the unique TCP flags combination. For example, if a flowlog consists of 3 packets with the flags [SYN, ACK, ACK] then the flowlog will have the 2-item list [SYN, ACK]. And if a flowlog consists of 3 pakcets with the flags [SYN+ACK, ACK, ACK] then the flowlog will have the 2-item list [SYN+ACK, ACK].

I have concerns with respect to the implementation on that. AFAIK, eBPF doesn't allow dynamic memory to be created. That would mean that, for each record, you'd need to create a big-enough, fixed-size array to store the flag sequences. I'm afraid this could impact in the overall memory and performance.

bpf/flows.c
```diff
@@ -171,7 +176,47 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
...
+    if ((void *)eth + sizeof(struct ethhdr) <= data_end) {
+        if (bpf_ntohs(eth->h_proto) == ETH_P_IP) {
+            const struct iphdr *iph = (struct iphdr *)(data + sizeof(struct ethhdr));
+            if ((void *)iph + sizeof(struct iphdr) <= data_end) {
+                if (iph->protocol == IPPROTO_TCP) {
+                    const struct tcphdr *th = (struct tcphdr *)(data + sizeof(struct iphdr));
+                    if ((void *)th + sizeof(struct tcphdr) <= data_end) {
+                        //bpf_printk("fin=%d, syn=%d, rst=%d, psh=%d, ack=%d, urg=%d, ece=%d, cwr=%d, res1=%d doff=%d\n",th->fin, th->syn, th->rst, th->psh, th->ack,th->urg,th->ece, th->cwr, th->res1, th->doff);
+                        if (th->fin) {
+                            flags_t = setBit(flags_t, 0);
```
mariomac, line 189: for speed, I'd just remove the `setBit` function and do the following:

```c
#define FIN_FLAG 0b1
#define SYN_FLAG 0b10
#define RST_FLAG 0b100
```

Then replace each invocation to `setBit` by:
```c
if (th->fin) {
   flags_t |= FIN_FLAG
}
if (th->syn) {
    flags_t |= SYN_FLAG
}
if (th->rst) {
    flags_t |= RST_FLAG
}
etc...
```

proto/flow.proto
mariomac: This is something critical to avoid that any update to a newer agent (without updating FLP) breaks the whole netobserv product.

mariomac (changes requested):

shach33:
I am working on it. Will push the changes soon.

mariomac:
@praveingk yes, a 16-bit number it's fine. Maybe I misunderstood @ronensc message and I thought some kind of list/array would be implemented to store that information.

ronensc:
> @praveingk yes, a 16-bit number it's fine. Maybe I misunderstood @ronensc message and I thought some kind of list/array would be implemented to store that information.

@mariomac you got me right. I wan't aware of eBPF limitations...

bpf/flows.c
```diff
@@ -42,6 +41,17 @@
...
+#define FIN_FLAG 0x1
+#define SYN_FLAG 0x2
+#define RST_FLAG 0x3
+#define URG_FLAG 0x4
+#define PSH_FLAG 0x5
+#define ECE_FLAG 0x6
+#define CWR_FLAG 0x7
+#define SYN_ACK_FLAG 0x8
+#define FIN_ACK_FLAG 0x9
+#define RST_ACK_FLAG 0xa
```
mariomac, line 53: Flags need to be powers of 2, so each one set their own exclusive bit
```suggestion
#define FIN_FLAG 1
#define SYN_FLAG 2
#define RST_FLAG 4
#define URG_FLAG 8
#define PSH_FLAG 16
#define ECE_FLAG 32
#define CWR_FLAG 64
#define SYN_ACK_FLAG 128
#define FIN_ACK_FLAG 256
#define RST_ACK_FLAG 512
```

pkg/exporter/kafka_proto_test.go
```diff
@@ -24,7 +24,7 @@ func IPAddrFromNetIP(netIP net.IP) flow.IPAddr {
 func TestProtoConversion(t *testing.T) {
 	wc := writerCapturer{}
 	kj := KafkaProto{Writer: &wc}
-	input := make(chan []*flow.Record, 10)
+	input := make(chan []*flow.Record, 11)
```
mariomac, line 27: I don't think it is needed, but I guess it neither hurts.

bpf/flows.c
```diff
@@ -171,7 +180,53 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
...
     id.direction = direction;
+    u16 flags_t = 0;
+    if ((void *)eth + sizeof(struct ethhdr) <= data_end) {
+        if (bpf_ntohs(eth->h_proto) == ETH_P_IP) {
+            const struct iphdr *iph = (struct iphdr *)(data + sizeof(struct ethhdr));
+            if ((void *)iph + sizeof(struct iphdr) <= data_end) {
+                if (iph->protocol == IPPROTO_TCP) {
+                    const struct tcphdr *th = (struct tcphdr *)(data + sizeof(struct iphdr));
+                    if ((void *)th + sizeof(struct tcphdr) <= data_end) {
+                        bpf_printk("fin=%d, syn=%d, rst=%d, psh=%d, ack=%d, urg=%d, ece=%d, cwr=%d, res1=%d doff=%d\n",th->fin, th->syn, th->rst, th->psh, th->ack,th->urg,th->ece, th->cwr, th->res1, th->doff);
```
mariomac, line 191: Are we sure we want to log a message for each packet? Apart of the overhead I'm afraid the logs will get flooded.

bpf/flows.c
```diff
@@ -171,7 +180,53 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
...
+			else if (th->urg) {
+                            flags_t |= URG_FLAG;
+                        }
+			else if (th->ece) {
+                            flags_t |= ECE_FLAG;
+                        }
+			else if (th->cwr) {
+                            flags_t |= CWR_FLAG;
+                        }
+                        bpf_printk("Flags: %d", flags_t);
```
mariomac, line 222: Same as before.

bpf/flows.c
```diff
@@ -198,9 +253,12 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
         // Key does not exist in the map, and will need to create a new entry.
         flow_metrics new_flow = {
             .packets = 1,
-            .bytes=skb->len,
+            .bytes = skb->len,
             .start_mono_time_ts = current_time,
             .end_mono_time_ts = current_time,
+            //Get flags from packet header
+            //Flags[0] = Fin, [1] = Syn, [2] = Rst, [3] = Psh, [4] = Ack, [5] = Urg, [6] = Ece, [7] = cwr
+            .flags = flags_t, //skb->len,
```
mariomac, line 261: I'd remove the `//skb->len` comment,  if we are not going to use it.

pkg/flow/record.go
```diff
@@ -65,8 +65,15 @@ type RecordMetrics struct {
...
-	Errno uint8
+	Flags           uint16
+	Errno           uint8
+
+	/*Fin   byte
+	Syn   byte
+	Rst   byte
+	Psh   byte
+	Ack   byte
+	Urg   byte*/
```
mariomac, line 76: I'd remove this.

praveingk:
@shach33 Can you rebase this PR to main branch since there are merge conflicts. After the rebase, I would like to add flags to ipfix export as well.

bpf/flows.c
```diff
@@ -171,7 +180,54 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
...
+			else if (th->cwr) {
+                            flags_t |= CWR_FLAG;
+                        }
+                    }
+                }
+            }
+        }
+    }
 
+    //struct tcphdr *tcp_header;
```
mariomac, line 230: If code it's unused, let's just remove it instead of commenting.

pkg/exporter/proto.go
```diff
@@ -2,13 +2,14 @@ package exporter
...
 	"encoding/binary"
-	"net"
-
 	"github.com/netobserv/netobserv-ebpf-agent/pkg/flow"
 	"github.com/netobserv/netobserv-ebpf-agent/pkg/pbflow"
 	"google.golang.org/protobuf/types/known/timestamppb"
+	"net"
 )
 
+//var klog = logrus.WithField("component", "exporter/KafkaProto")
```
mariomac, line 11: Please remove line instead of commenting.

mariomac (changes requested):
Much better, thanks! Few minor comments.

bpf/flows.c
```diff
@@ -97,7 +138,7 @@ static inline int fill_iphdr(struct iphdr *ip, void *data_end, flow_id *id) {
 }
 
 // sets flow fields from IPv6 header information
-static inline int fill_ip6hdr(struct ipv6hdr *ip, void *data_end, flow_id *id) {
+static inline int fill_ip6hdr(struct ipv6hdr *ip, void *data_end, flow_id *id, u16 *flags) {
```
jotak, line 141: shouldn't we call `set_flags`  here too, in `case IPPROTO_TCP` block?

mariomac: Good catch! We always forget to test IPv6 :-(

jotak (approved):
lgtm!

mariomac (approved):

jotak:
Merging!
Kudos @shach33 @praveingk

PR #69, opened 2022-11-24
Fix e2e tests

mariomac:


praveingk (approved):
Works now with the changes 👍🏽

PR #78, opened 2022-12-19
NETOBSERV-759: decorate flows with agent IP

mariomac:
Only for Protocol Buffers encoding. IPFIX is still missing.

This PR creates a new "Decorator" instance and moves there also the interface naming process.

At least for OVN-K, the default `external` IP parsing coincides with the Node IP without any kind of tuning from the NOO:
image


eranra:
@mariomac can you provide some motivation for this enrichment :-) --- we have the network Interface ID --- can't we get that data from k8s and not send that field over and over for each flow?

jotak:
@eranra do you have strong concerns about this?
It's really solving a lot of issues that we have, regarding the interpretation of `FlowDirection` (what we assume on frontend side != what it is at the low level, and knowing the Agent IP was the missing piece between them)

jotak:
tested and works as expected
@praveingk do you want to review? You know better the code than me :)

pkg/exporter/ipfix.go
```diff
@@ -12,6 +12,9 @@ import (
 
 var ilog = logrus.WithField("component", "exporter/IPFIXProto")
 
+// TODO: encode also the equivalent of the Protobuf's AgentIP field in a format that is binary-
```
jotak, line 15: For information, in goflow's IPFIX this was called "SamplerAddress": https://github.com/cloudflare/goflow#output-format

In IPFIX spec (https://www.iana.org/assignments/ipfix/ipfix.xhtml) this is `exporterIPv4Address` / `exporterIPv6Address`

pkg/flow/decorator.go
```diff
@@ -0,0 +1,22 @@
+package flow
+
+import (
+	"net"
+)
+
+type InterfaceNamer func(ifIndex int) string
+
+// Decorate adds to the flows extra metadata fields that are not directly fetched by eBPF:
+// - The interface name (corresponding to the interface index in the flow).
```
jotak, line 10: @mariomac the interface name was already in the flows, is it just that you refactored at the same time the way it is added, right? Or did you change the flow content wrt interface?

jotak (approved):
thanks!

pkg/flow/decorator.go
mariomac: Yes, it was added in an earlier stage in the flow creation, but the constructor required to inject an external component (interfaceNamer) and I think that moving the extra decoration to another stage makes the code cleaner.

mariomac:
Merging to unblock other PRs/tests and, if @praveingk has any concern about the current PR, we can address them into another PR.

pkg/exporter/ipfix.go
praveingk: @jotak the ExportingProcess has a net.Conn, which could be used to get the info. Yes, exporterIPv4(/6)Address seems right though

praveingk: However, the ExportingProcess could be used only if the AgentIPIface is set as "External"

PR #74, opened 2022-12-15
NETOBSERV-755: ignore flow direction in deduplication

mariomac:
Being also backported to 4.12: https://github.com/netobserv/netobserv-ebpf-agent/pull/75

This is technically a breaking change, as some flows that were marked as "Duplicate: false" will be now marked as "Duplicate: true".

From the Netobserv user experience, this shouldn't have any visible impact as long as they don't set the `agent.debug.env.DEDUPER_JUST_MARK=false` flag.

It should have the positive impact of more accurate metric aggregations.

PR #79, opened 2022-12-22
Revert "Revert "NETOBSERV-755: ignore flow direction in deduplication…

jotak:
… (#74)" (#77)"

This reverts commit 0e3f5eac6083b888f6d07797c90e574298c1b86b.

### DIFF
--- a/pkg/flow/deduper.go
+++ b/pkg/flow/deduper.go
@@ -67,6 +67,7 @@ func (c *deduperCache) isDupe(key *RecordKey) bool {
 	// zeroes fields from key that should be ignored from the flow comparison
 	rk.IFIndex = 0
 	rk.DataLink = DataLink{}
+	rk.Direction = 0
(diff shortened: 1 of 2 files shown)


mariomac (approved):

jotak:
@mariomac can we merge it? (some checks are stuck in progress)

mariomac:
@jotak yes, we can. The 1.17 tests were removed because we updated the source to 1.18 but by any reason I don't understand, they are still there but they don't run. I hoped to fix it with this but I can't: https://github.com/netobserv/netobserv-ebpf-agent/pull/76

jotak:
Ok, I've found the culprit: it's the branch protection rules that prohibit merge if some status checks don't pass, but sadly, the status checks listed in the protection rules **include the startegy version** so it was still configured on "test (1.17)" or "e2e tests 1.17" 🤪

For now I've removed this protection ; checking on FLP, we don't use protection on status check either

PR #81, opened 2023-01-02
NETOBSERV-617: split big payloads in GRPC exporter

mariomac:
Only for GRPC exporter, payloads with more than 10000 flows (default configurable value) are split in multiple messages that are submitted sequentially.

### DIFF
--- a/pkg/exporter/proto.go
+++ b/pkg/exporter/proto.go
@@ -11,14 +11,21 @@ import (
 
 // flowsToPB is an auxiliary function to convert flow records, as returned by the eBPF agent,
 // into protobuf-encoded messages ready to be sent to the collector via GRPC
-func flowsToPB(inputRecords []*flow.Record) *pbflow.Records {
+func flowsToPB(inputRecords []*flow.Record, maxLen int) []*pbflow.Records {
 	entries := make([]*pbflow.Record, 0, len(inputRecords))
 	for _, record := range inputRecords {
 		entries = append(entries, flowToPB(record))
 	}
-	return &pbflow.Records{
-		Entries: entries,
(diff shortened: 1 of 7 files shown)


pkg/exporter/proto.go
```diff
@@ -11,14 +11,21 @@ import (
...
 // into protobuf-encoded messages ready to be sent to the collector via GRPC
-func flowsToPB(inputRecords []*flow.Record) *pbflow.Records {
+func flowsToPB(inputRecords []*flow.Record, maxLen int) []*pbflow.Records {
 	entries := make([]*pbflow.Record, 0, len(inputRecords))
 	for _, record := range inputRecords {
 		entries = append(entries, flowToPB(record))
 	}
-	return &pbflow.Records{
-		Entries: entries,
+	var records []*pbflow.Records
```
jotak, line 19: Could it be better done in one pass rather than 2 passes?

jotak (approved):
lgtm, just wondering if a small perf improvement is possible when creating flow slices

pkg/exporter/proto.go
mariomac: I'll do it despite I'm afraid it will look a bit less readable. Also consider that in the second step we aren't copying arrays but just creating references to subslices pointing to the data in the original "entries" slice, so the impact in performance shouldn't be really high.
image
But anyway let me do the change and then decide which one do you prefer.

jotak: no no, that's fine, you can leave it as it is
I though it involved copying slices

jotak: thanks btw for the info!

PR #84, opened 2023-01-03
ghactions: remove strategy matrix

jotak:
The goal is to allow branch protection rule with these status checks, but we don't want to have to modify every time we update the go version (as they're referred to with their go version included when startegy matrix is used)

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -10,9 +10,6 @@ jobs:
   test:
     name: test
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


jotak:
@mariomac as expected, if I remove the matrix, then I can configure the requirements for merge without the go version inferred: https://github.com/netobserv/netobserv-ebpf-agent/settings/branch_protection_rules

mariomac (approved):

mariomac:
Amazing! Thanks for handling this

PR #88, opened 2023-01-17
NETOBSERV-661: fix crash on shutdown

mariomac:
The flush operation was done asynchronously in another goroutine so it was frequent that, when the agent is shutdown, the destination channel was closed in the main goroutine while the flush goroutine was performing the channel submission.

This PR verifies, before channel send, that the context of the main goroutine is not closed.

This PR also removes some warnings on shutdown due to the agent trying to close unexisting resources of FLP and the Console Plugin, when they are removed just before the agent shutdown.

### DIFF
--- a/pkg/ebpf/tracer.go
+++ b/pkg/ebpf/tracer.go
@@ -279,8 +279,9 @@ func (m *FlowFetcher) Close() error {
 func doIgnoreNoDev[T any](sysCall func(T) error, dev T, log *logrus.Entry) error {
 	if err := sysCall(dev); err != nil {
 		if errors.Is(err, unix.ENODEV) {
-			log.WithError(err).Debug("can't delete. Probably the " +
-				"associated container or interface has been deleted immediatelly before this task")
+			log.WithError(err).Error("can't delete. Ignore this error if other pods or interfaces " +
+				" are also being deleted at this moment. For example, if you are undeploying " +
+				" a FlowCollector or Deployment where this agent is part of")
 		} else {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


OlivierCazade (approved):
LGTM

pkg/ebpf/tracer.go
```diff
@@ -264,6 +267,27 @@ func (m *FlowFetcher) Close() error {
...
+// undeploying the FlowCollector could cause the agent to try to remove resources
+// from Pods that have been removed immediately before (e.g. flowlogs-pipeline or the
+// console plugin), so we avoid logging some errors that would unnecessarily raise the
+// user's attention.
+// This function uses generics because the set of provided functions accept different argument
+// types.
+func doIgnoreNoDev[T any](sysCall func(T) error, dev T, log *logrus.Entry) error {
+	if err := sysCall(dev); err != nil {
+		if errors.Is(err, unix.ENODEV) {
+			log.WithError(err).Debug("can't delete. Probably the " +
```
jotak, line 282: Maybe it should be still logged as Error level, but with a message like "blah blah - ignore this error if the container or interface is intentionally being deleted, such as on pod shutdown" or something similar?
My concern is that we may miss legitimate errors.

jotak:
Small comment, lgtm otherwise

pkg/ebpf/tracer.go
mariomac: Yeah good appreciation. Changing it.

jotak (approved):
lgtm!

Issue #93, opened 2023-02-10
IPFIX export for v6 records fails


praveingk:
Got the below error when exporting IPFIX records for v6 packets

```
ERRO[0010] Failed in send IPFIX data record              collector="localhost:9999" component=exporter/IPFIXProto error="error when doing sanity check:process: Data Record does not pass the min required length (93) check for template ID 257"
ERRO[0010] Failed in send IPFIX data record              
```

PR #94, opened 2023-02-10
Fix v6 ipfix export

praveingk:
This PR fixes https://github.com/netobserv/netobserv-ebpf-agent/issues/93

### DIFF
--- a/pkg/exporter/ipfix.go
+++ b/pkg/exporter/ipfix.go
@@ -319,9 +319,16 @@ func (ipf *IPFIX) sendDataRecord(log *logrus.Entry, record *flow.Record, v6 bool
 	if err != nil {
 		return err
 	}
-	err = dataSet.AddRecord(ipf.entitiesV4, templateID)
-	if err != nil {
-		return err
+	if v6 {
+		err = dataSet.AddRecord(ipf.entitiesV6, templateID)
+		if err != nil {
+			return err
+		}
+	} else {
+		err = dataSet.AddRecord(ipf.entitiesV4, templateID)
+		if err != nil {
+			return err
+		}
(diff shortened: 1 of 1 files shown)


pkg/exporter/ipfix.go
```diff
@@ -319,9 +319,16 @@ func (ipf *IPFIX) sendDataRecord(log *logrus.Entry, record *flow.Record, v6 bool
 	if err != nil {
 		return err
 	}
-	err = dataSet.AddRecord(ipf.entitiesV4, templateID)
-	if err != nil {
-		return err
+	if v6 {
```
msherif1234, line 322: its good practice to avoid if() {
} else {
}

praveingk: @msherif1234 Wouldn't that perform `data.Set.AddRecord` multiple times for v6?

jotak: +1 @praveingk 
I think the good practice referred to is writing `if { ...return } ... return` instead of `if { ...return } else { ...return}`, but it doesn't apply here as not all paths return in the conditional expressions

msherif1234: yeah @praveingk I was commenting in general and yes whatu have make sense
/lgtm

praveingk: I see :)

msherif1234: Hi @praveingk 
so we can do this
```golang
entities := ipf.entitiesV4
if v6 {
  entities = ipf.entitiesV6
}
if err := dataSet.AddRecord(entities, templateID); err != nil {
  return err
}
```

PR #96, opened 2023-02-13
Move to go1.19

jotak:


jotak:
Moving away from the go-toolset image as it often lays some versions behind

Dockerfile
```diff
@@ -1,8 +1,7 @@
 # Build the manager binary
-FROM registry.access.redhat.com/ubi9/go-toolset:1.18.4 as builder
+FROM docker.io/library/golang:1.19 as builder
```
msherif1234, line 2: is this intentional ? we should find and go1.19 from redhat registry right ?

msherif1234 (commented):
pls restore changes to bpf_bpfel/g.go files no need to merge them
I have a question about new builder image

msherif1234:
> Moving away from the go-toolset image as it often lays some versions behind

is that considered secure from RH prespective ? did u check to see if there are other alternatives ?

Dockerfile
jotak: yes it is intentional, ubi's go-toolset often lies behind in terms of go version so we want something more up to date.
This is for the upstream/community build, so using `golang:1.19` is fine (we actually do the same already for a while in flowlogs-pipeline); the Dockerfile for downstream is different and uses images from redhat registry.

jotak:
> pls restore changes to bpf_bpfel/g.go files no need to merge them I have a question about new builder image

it wasn't a manual edit, it's a new format from `go fmt` (when using go 1.19), so even though they are generated, I guess they would still adopt that format if re-generated

PR #98, opened 2023-02-13
NETOBSERV-889: fix too many colons in address error

msherif1234:
This PR fixed two issues

1- when hostIP is IPv6 address we endup with following error when gRPC try to connect
```bash
collector="2604:1380:4642:7e00::5b:2055" component=exporter/GRPCProto error="rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing dial tcp: address 2604:1380:4642:7e00::5b:2055: too many colons in address\""
```

for IPv6 we need to use `[2604:1380:4642:7e00::5b]:2055` format.

2- on single stack ipv6 cluster with IPType=="any" , ipv4 DNS address was incorrectly used which prevented ebpf agent to come up and the pod stay in crashloop with this error:-
```bash
time="2023-02-16T21:06:17Z" level=fatal msg="can't instantiate NetObserv eBPF Agent" error="acquiring Agent IP: can't establish an external connection dial udp 8.8.8.8:80: connect: network is unreachable"
```
Signed-off-by: msherif1234 <mmahmoud@redhat.com>

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -192,8 +192,7 @@ func buildFlowExporter(cfg *Config) (node.TerminalFunc[[]*flow.Record], error) {
 			return nil, fmt.Errorf("missing target host or port: %s:%d",
 				cfg.TargetHost, cfg.TargetPort)
 		}
-		target := fmt.Sprintf("%s:%d", cfg.TargetHost, cfg.TargetPort)
-		grpcExporter, err := exporter.StartGRPCProto(target, cfg.GRPCMessageMaxFlows)
+		grpcExporter, err := exporter.StartGRPCProto(cfg.TargetHost, cfg.TargetPort, cfg.GRPCMessageMaxFlows)
 		if err != nil {
 			return nil, err
 		}
@@ -243,9 +242,7 @@ func buildFlowExporter(cfg *Config) (node.TerminalFunc[[]*flow.Record], error) {
 			return nil, fmt.Errorf("missing target host or port: %s:%d",
 				cfg.TargetHost, cfg.TargetPort)
 		}
-		target := fmt.Sprintf("%s:%d", cfg.TargetHost, cfg.TargetPort)
-
-		ipfix, err := exporter.StartIPFIXExporter(target, "udp")
+		ipfix, err := exporter.StartIPFIXExporter(cfg.TargetHost, cfg.TargetPort, "udp")
 		if err != nil {
 			return nil, err
 		}
@@ -255,9 +252,7 @@ func buildFlowExporter(cfg *Config) (node.TerminalFunc[[]*flow.Record], error) {
 			return nil, fmt.Errorf("missing target host or port: %s:%d",
 				cfg.TargetHost, cfg.TargetPort)
 		}
-		target := fmt.Sprintf("%s:%d", cfg.TargetHost, cfg.TargetPort)
-
-		ipfix, err := exporter.StartIPFIXExporter(target, "tcp")
+		ipfix, err := exporter.StartIPFIXExporter(cfg.TargetHost, cfg.TargetPort, "tcp")
 		if err != nil {
 			return nil, err
 		}
--- a/pkg/exporter/grpc_proto.go
+++ b/pkg/exporter/grpc_proto.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/netobserv/netobserv-ebpf-agent/pkg/flow"
 	"github.com/netobserv/netobserv-ebpf-agent/pkg/grpc"
+	"github.com/netobserv/netobserv-ebpf-agent/pkg/utils"
 	"github.com/sirupsen/logrus"
 )
 
@@ -14,7 +15,8 @@ var glog = logrus.WithField("component", "exporter/GRPCProto")
 // by its input channel, converts them to *pbflow.Records instances, and submits
 // them to the collector.
 type GRPCProto struct {
-	hostPort   string
+	hostIP     string
+	hostPort   int
 	clientConn *grpc.ClientConnection
 	// maxFlowsPerMessage limits the maximum number of flows per GRPC message.
 	// If a message contains more flows than this number, the GRPC message will be split into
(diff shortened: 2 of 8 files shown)


Amoghrd:
@msherif1234 Tried updating the CSV with this eBPF image, the eBPF pods went to CrashLoopBackoff with error
`time="2023-02-13T21:23:28Z" level=info msg="starting NetObserv eBPF Agent"
time="2023-02-13T21:23:28Z" level=info msg="initializing Flows agent" component=agent.Flows
time="2023-02-13T21:23:28Z" level=fatal msg="can't instantiate NetObserv eBPF Agent" error="acquiring Agent IP: can't establish an external connection dial udp 8.8.8.8:80: connect: network is unreachable"`

pkg/exporter/grpc_proto.go
```diff
@@ -37,7 +40,8 @@ func StartGRPCProto(hostPort string, maxFlowsPerMessage int) (*GRPCProto, error)
 // ExportFlows accepts slices of *flow.Record by its input channel, converts them
 // to *pbflow.Records instances, and submits them to the collector.
 func (g *GRPCProto) ExportFlows(input <-chan []*flow.Record) {
-	log := glog.WithField("collector", g.hostPort)
+	grpcSocket := fmt.Sprintf("[%s]:%d", g.hostIP, g.hostPort)
```
stleerh, line 43: Square brackets should only be added if g.hostIP is an IPv6 address, otherwise it should just be "%s:%d".

pkg/grpc/client.go
```diff
@@ -13,9 +15,10 @@ type ClientConnection struct {
 	conn   *grpc.ClientConn
 }
 
-func ConnectClient(address string) (*ClientConnection, error) {
+func ConnectClient(address string, port int) (*ClientConnection, error) {
 	// TODO: allow configuring some options (keepalive, backoff...)
-	conn, err := grpc.Dial(address,
+	grpcSocket := fmt.Sprintf("[%s]:%d", address, port)
```
stleerh, line 20: Square brackets should only be added if g.hostIP is an IPv6 address, otherwise it should just be "%s:%d".

pkg/agent/agent.go
```diff
@@ -192,8 +192,7 @@ func buildFlowExporter(cfg *Config) (node.TerminalFunc[[]*flow.Record], error) {
 			return nil, fmt.Errorf("missing target host or port: %s:%d",
 				cfg.TargetHost, cfg.TargetPort)
 		}
-		target := fmt.Sprintf("%s:%d", cfg.TargetHost, cfg.TargetPort)
-		grpcExporter, err := exporter.StartGRPCProto(target, cfg.GRPCMessageMaxFlows)
+		grpcExporter, err := exporter.StartGRPCProto(cfg.TargetHost, cfg.TargetPort, cfg.GRPCMessageMaxFlows)
```
stleerh, line 195: There are two other places where this needs to be done. See lines 245 and 257 in this file.

pkg/exporter/grpc_proto_test.go
```diff
@@ -71,7 +70,7 @@ func TestGRPCProto_SplitLargeMessages(t *testing.T) {
 
 	const msgMaxLen = 10000
 	// Start GRPCProto exporter stage
-	exporter, err := StartGRPCProto(fmt.Sprintf("127.0.0.1:%d", port), msgMaxLen)
+	exporter, err := StartGRPCProto("127.0.0.1", port, msgMaxLen)
```
stleerh, line 73: How about add a IPv6 test case (e.g. "::1") here and in grpc_test.go.

Amoghrd:
I get the same above error with source install too; after updating the eBPF image in deployments

pkg/exporter/grpc_proto.go
msherif1234: this one is just debug log not used to opne grpc channel so I can use %s:%d for both

pkg/grpc/client.go
msherif1234: I thought [] should work for both AF ? anyway I will make it AF specific

pkg/agent/agent.go
msherif1234: hmm I changed only gRPC do we need to make it for IPFIX too ? probably good idea will change those too

jotak (dismissed):
/lgtm

pkg/utils/utils.go
```diff
@@ -0,0 +1,16 @@
...
+	"fmt"
+	"net"
+)
+
+// GetSocket returns socket string in the correct format based on address family
+func GetSocket(hostIP string, hostPort int) string {
+	socket := fmt.Sprintf("%s:%d", hostIP, hostPort)
+	ipAddr := net.ParseIP(hostIP)
+	if ipAddr.To4() == nil {
+		socket = fmt.Sprintf("[%s]:%d", hostIP, hostPort)
```
stleerh, line 13: (Minor) The only minor issue with this is that if it is any non IPv4 address, such as "1.2.3" or "example.com", the socket string will have square brackets.  
  
Ideally, the preferred behavior is if it's an IPv6 address, put square brackets, otherwise, leave it as-is.  That way, the error message for "1.2.3" with port 2055 won't be that \[1.2.3\]:2055 failed.  Unfortunately, there is no function in IP package to check for IPv6, so you have to resort to something like it has a colon character (which I know some people don't like).  Nevertheless, it's minor so I'm okay going with this solution.

msherif1234: there is `To16` in net package but it really doesn't differentiate between IPv4 and IPv6 I ran into uses using it in the pass here is an example where it shows its unexpected behavior https://go.dev/play/p/VRP6XKdfgJy 
so the reality is this hostIP so it can't be malformed format we will have way many issues outside no if that ever happened, hence I used that route

msherif1234: fixed it found away so malformed v4 address won't show up as ipv6

stleerh: I like the solution!

stleerh (approved):

memodi:
> @msherif1234 Tried updating the CSV with this eBPF image, the eBPF pods went to CrashLoopBackoff with error `time="2023-02-13T21:23:28Z" level=info msg="starting NetObserv eBPF Agent" time="2023-02-13T21:23:28Z" level=info msg="initializing Flows agent" component=agent.Flows time="2023-02-13T21:23:28Z" level=fatal msg="can't instantiate NetObserv eBPF Agent" error="acquiring Agent IP: can't establish an external connection dial udp 8.8.8.8:80: connect: network is unreachable"`

tried this more and discussed with error, our initial understanding of not being able to pull image: quay.io/netobserv/netobserv-ebpf-agent:ceaa4ca from quay was incorrect, it was able to pull Image and landing in above **runtime** error. This is likely a bug with new code.

memodi:
/qe-approved
Verified fix works for both IPv6 single stack and IPv4 clusters as well, thanks!

memodi:
@jotak - is something wrong with labels? I don't see qe-approved label being added.

OlivierCazade (approved):
LGTM

jotak:
thanks @msherif1234 !
/approve

PR #100, opened 2023-03-02
NETOBSERV-848 eBPF agent multi-arch builds (upstream)

jpinsonneau:
Implements multi arch builds and add documentation and consistency

```
$ make help

Usage:
  make <target>

General
  help                  Display this help.
  vendors               Check go vendors
  prereqs               Check if prerequisites are met, and install missing dependencies

Develop
  fmt                   Run go fmt against code.
  lint                  Lint the code
  generate              Generate artifacts of the code repo (pkg/ebpf and pkg/proto packages)
  docker-generate       Create the container that generates the eBPF binaries
  compile               Compile ebpf agent project
  test                  Test code using go test
  coverage-report       Generate coverage report
  coverage-report-html  Generate HTML coverage report
  tests-e2e             Run e2e tests

Images
  image-build           Build MULTIARCH_TARGETS images
  image-push            Push MULTIARCH_TARGETS images
  manifest-build        Build MULTIARCH_TARGETS manifest
  manifest-push         Push MULTIARCH_TARGETS manifest
  ci-images-build       Build CI images
  ci-images-push        Push CI images

shortcuts helpers
  build                 Test and Build ebpf agent
  build-image           Build MULTIARCH_TARGETS images
  push-image            Push MULTIARCH_TARGETS images
  build-manifest        Build MULTIARCH_TARGETS manifest
  push-manifest         Push MULTIARCH_TARGETS manifest
  images                Build and push MULTIARCH_TARGETS images and related manifest
  build-ci-images       Build CI images
  push-ci-images        Push CI images
  ci-images             Build and push CI images

```

image

Please use https://github.com/netobserv/flowlogs-pipeline/pull/426 for main discussion and this PR only for specific items

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,11 +140,13 @@ coverage-report-html: cov-exclude-generated
 
 .PHONY: image-build
 image-build:
+	trap 'exit' INT
 	$(foreach target,$(MULTIARCH_TARGETS),$(call build_target,$(target)))
 	$(call build_manifest)
 
 .PHONY: image-push
 image-push: image-build
+	trap 'exit' INT
 	$(foreach target,$(MULTIARCH_TARGETS),$(call push_target,$(target)))
 	$(call push_manifest)
 
--- a/Makefile
+++ b/Makefile
@@ -1,86 +1,95 @@
-# SW_VERSION defines the project version for the bundle.
+# VERSION defines the project version for the bundle.
 # Update this value when you upgrade the version of your project.
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
-# - use the SW_VERSION as arg of the bundle target (e.g make bundle SW_VERSION=0.0.2)
-# - use environment variables to overwrite this value (e.g export SW_VERSION=0.0.2)
-SW_VERSION ?= main
-BUILD_VERSION := $(shell git describe --long HEAD)
+# - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
+# - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
+VERSION ?= main
 BUILD_DATE := $(shell date +%Y-%m-%d\ %H:%M)
+TAG_COMMIT := $(shell git rev-list --abbrev-commit --tags --max-count=1)
+TAG := $(shell git describe --abbrev=0 --tags ${TAG_COMMIT} 2>/dev/null || true)
 BUILD_SHA := $(shell git rev-parse --short HEAD)
+BUILD_VERSION := $(TAG:v%=%)
+ifneq ($(COMMIT), $(TAG_COMMIT))
+	BUILD_VERSION := $(BUILD_VERSION)-$(BUILD_SHA)
+endif
+ifneq ($(shell git status --porcelain),)
+	BUILD_VERSION := $(BUILD_VERSION)-dirty
+endif
 
+# Go architecture and targets images to build
 GOARCH ?= amd64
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -33,13 +35,8 @@ PROTOC_ARTIFACTS := pkg/pbflow
...
 
 # Image building tool (docker / podman)
-ifndef OCI_BIN
-	ifeq (,$(shell which podman 2>/dev/null))
-	OCI_BIN=docker
-	else
-	OCI_BIN=podman
-	endif
-endif
+OCI_BIN_PATH = $(shell which podman  || which docker)
```
stleerh, line 38: *   If podman doesn't exist, for example, it will display an error message.  This should be suppressed. 
*   It should also check that one of podman or docker exists and quit if that is not the case.
*   Use := instead of =.  
      
Suggest:  
```
OCI_BIN_PATH := $(shell which podman 2> /dev/null || which docker 2> /dev/null)
ifeq ($(OCI_BIN_PATH),)
$(error Requires podman or docker in PATH)
endif
```

jpinsonneau: Yes, I relied on [FLP Makefile](https://github.com/netobserv/flowlogs-pipeline/blob/main/Makefile#L17-L18) for such changes. Let's all agree on how to manage that and apply it everywhere !

@jotak @OlivierCazade  @msherif1234

Makefile
```diff
@@ -115,14 +112,43 @@ coverage-report-html: cov-exclude-generated
 image-build: ## Build OCI image with the manager.
 	$(OCI_BIN) build --build-arg SW_VERSION="$(SW_VERSION)" -t ${IMG} .
```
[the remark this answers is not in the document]
jpinsonneau: so the var will update it's value one after the other for multiarch build ? you will loose the ability to run in parallel like `make -j multiarch-manifest-build` :thinking:

Makefile
jotak: How about a slight variation:

```
OCI_BIN_PATH ?= $(shell which podman 2> /dev/null || which docker 2> /dev/null)
ifeq ($(OCI_BIN_PATH),)
$(error Requires OCI-compliant CLI such as podman or docker in PATH)
endif
```

`?=` allows for overriding it, e.g. using a different installation path or another OCI compliant tool

Makefile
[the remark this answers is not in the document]
msherif1234: I was thinking of two make invocation in case of amd and ppc tomorrow could be another platform for ibm or windows

Makefile
stleerh: The only problem is that in other parts of the Makefile, it generally uses OCI\_BIN, which is the basename of OCI\_BIN\_PATH (either 'podman' or 'docker') instead of OCI\_BIN\_PATH which has the absolute path.  Therefore, it's possible if you override OCI\_BIN\_PATH, OCI\_BIN could fail or even use a different version depending on PATH.  
  
You could change all OCI\_BIN usage to OCI\_BIN\_PATH.  There is at least one case where it's using OCI\_BIN to determine if you are running docker or podman.

jpinsonneau: your suggestions doesn't work with `docker.io/library/golang:1.19` image + it's common to simply use `$(shell which podman  || which docker)` without further checks

I'll keep this as is for now

Makefile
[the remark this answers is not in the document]
jpinsonneau: https://github.com/netobserv/netobserv-ebpf-agent/pull/100/commits/d54d2218834ced8d8da5cece60eda50ff93f5d56

PR #103, opened 2023-03-15
NETOBSERV-934: Add support for SCTP, ICMPv4/v6 protocols to ebpf agent

msherif1234:
Extend L4 protocols support to include
- SCTP
- ICMPv4
- ICMPv6

Note: SCTP structure is not exported hence I am using local version of it in the ebpf code I checked with RHEL team and they said this probably fine because its very unlikely this structure will ever change and there is no plans to exported either.

unit-testing
```shell
ipv4: 07:58:14.541222 enp1s0 IP 192.168.122.69:0 > 192.168.122.1:0: protocol:icmp type: 8 code: 0 dir:1 bytes:490 packets:5 flags:0 ends: 07:58:18.637230
ipv4: 07:58:14.541512 enp1s0 IP 192.168.122.1:0 > 192.168.122.69:0: protocol:icmp type: 0 code: 0 dir:0 bytes:490 packets:5 flags:0 ends: 07:58:18.637502
```


go.mod
```diff
@@ -21,6 +21,7 @@ require (
 	k8s.io/api v0.24.0
 	k8s.io/apimachinery v0.24.0
 	k8s.io/client-go v0.24.0
+	k8s.io/klog/v2 v2.60.1
```
jotak, line 24: logrus is used everywhere else in this project, is there a reason to introduce klog here instead?

msherif1234: No will switch

bpf/flows.c
```diff
@@ -55,6 +54,13 @@
 #define FIN_ACK_FLAG 0x200
 #define RST_ACK_FLAG 0x400
 
+struct sctphdr {
```
praveingk, line 57: Could move this declaration be moved to flow.h

praveingk (dismissed):
Looks good 👍🏽

bpf/flows.c
msherif1234: this structure used only internally by flow.c not need to be exported outside

bpf/flows.c
```diff
@@ -104,71 +112,107 @@ static inline void set_flags(struct tcphdr *th, u16 *flags) {
...
+    u16 dst_port;
+    // ICMPv4/ICMPv6 type value
+    u8 icmp_type;
+    // ICMPv4/ICMPv6 code value
+    u8 icmp_code;
+};
+
+// Extract L4 info for the supported protocols
+static inline void fill_l4info(void *l4_hdr_start, void *data_end, u8 protocol,
+                               u16 *flags, struct l4_info_t *l4_info) {
```
praveingk, line 130: Should we move flags also to l4_info ? since its technically l4 :)

msherif1234: done

praveingk (approved):

PR #107, opened 2023-04-04
NETOBSERV-926: Added key to kafka message for connection tracking consistency

OlivierCazade:
Add keys to each kafka message. Kafka then ensure that all messages with the same key will be consumed by the same client.

We use src and destination IP as key so all messages from a connection will be consumed by the same transformer so connection tracking metrics will be accurate.

### DIFF
--- a/pkg/exporter/kafka_proto.go
+++ b/pkg/exporter/kafka_proto.go
@@ -28,6 +28,17 @@ func (kp *KafkaProto) ExportFlows(input <-chan []*flow.Record) {
 	}
 }
 
+func getFlowKey(record *flow.Record) []byte {
+	for k := range record.Id.SrcIp {
+		if record.Id.SrcIp[k] < record.Id.DstIp[k] {
+			return append(record.Id.SrcIp[:], record.Id.DstIp[:]...)
+		} else if record.Id.SrcIp[k] > record.Id.DstIp[k] {
+			return append(record.Id.DstIp[:], record.Id.SrcIp[:]...)
+		}
+	}
+	return append(record.Id.SrcIp[:], record.Id.DstIp[:]...)
+}
+
 func (kp *KafkaProto) batchAndSubmit(records []*flow.Record) {
 	klog.Debugf("sending %d records", len(records))
 	msgs := make([]kafkago.Message, 0, len(records))
@@ -37,7 +48,7 @@ func (kp *KafkaProto) batchAndSubmit(records []*flow.Record) {
 			klog.WithError(err).Debug("can't encode protobuf message. Ignoring")
 			continue
 		}
-		msgs = append(msgs, kafkago.Message{Value: pbBytes})
+		msgs = append(msgs, kafkago.Message{Value: pbBytes, Key: getFlowKey(record)})
 	}
 
 	if err := kp.Writer.WriteMessages(context.TODO(), msgs...); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/exporter/kafka_proto.go
```diff
@@ -28,6 +29,26 @@ func (kp *KafkaProto) ExportFlows(input <-chan []*flow.Record) {
 	}
 }
 
+func getIPKey(ip [16]uint8) string {
```
msherif1234, line 32: nit: should we move this util to record.go where we keep other tools for conversion ?
also I didn't take a close look at net package to see if there is another ways to doing this but I assume u checked ?

msherif1234 (commented):
pls add to the PR the description to the issue and how this PR fix it too Thanks!!

pkg/exporter/kafka_proto.go
OlivierCazade: I realized that are transforming from byte array to string and then back to byte array.

I reworked the PR to directly work with byte arrays.

OlivierCazade:
> pls add to the PR the description to the issue and how this PR fix it too Thanks!!

Done, thanks.

pkg/exporter/kafka_proto.go
```diff
@@ -28,6 +28,17 @@ func (kp *KafkaProto) ExportFlows(input <-chan []*flow.Record) {
 	}
 }
 
+func getFlowKey(record *flow.Record) []byte {
+	for k := range record.Id.SrcIp {
+		if record.Id.SrcIp[k] < record.Id.DstIp[k] {
+			return append(record.Id.SrcIp[:], record.Id.DstIp[:]...)
+		} else if record.Id.SrcIp[k] > record.Id.DstIp[k] {
```
msherif1234, line 35: maybe just check for this condition since `==` and `<` both appends the same so we can avoid if else ?

OlivierCazade: No, if both bytes are equal we first want to iterate another time. But we return as soon as there is a difference.

pkg/exporter/kafka_proto.go
```diff
@@ -28,6 +28,17 @@ func (kp *KafkaProto) ExportFlows(input <-chan []*flow.Record) {
 	}
 }
 
+func getFlowKey(record *flow.Record) []byte {
+	for k := range record.Id.SrcIp {
```
jotak, line 32: this assumes number of bytes in src and dest are equal .. what if there's an ipv4 and an ipv6 ?

jotak: maybe check `len` first to return the smallest first (or last) if they aren't equal?
Also, can you add a comment about why we need to sort this? I fear if we look at the code later we might hav forgotten :-)

OlivierCazade: SrcIp and DstIp are not slice but fixed length array:
https://github.com/netobserv/netobserv-ebpf-agent/blob/c62173ac9558581fee03e63955a7eb8941178d05/pkg/ebpf/bpf_bpfel.go#L23

And more general question, is it possible to have a src ipv4 and a dst ipv6? I would have said no but you made me wonder

msherif1234: that is probably impossible to have a pkt header "flow" with src and dst from different address family.
/lgtm

pkg/exporter/kafka_proto.go
msherif1234: ok Thanks for the explaining it

PR #111, opened 2023-04-13
Filter interfaces by their associated IPs

masonj5n:
We have a use case where it's easy for us to pass the IP(s) of the interfaces we'd like to monitor as an environment variable but quite a bit harder to know the names of those interfaces in advance.

Here's a first pass at adding another filter type that allows/disallows interfaces based on whether it's associated with the passed in IPs. There's still some logic around the poller/watcher for interfaces that'd need some additions to also watch for when IPs are attached/unattached and I'm sure other things that would need updates too.

Is there any interest in adding a feature like this? Happy to change approach from what I have here or otherwise do what still needs done to get it merged if so!

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -69,16 +69,16 @@ require (
 	github.com/prometheus/common v0.32.1 // indirect
 	github.com/prometheus/procfs v0.7.3 // indirect
 	github.com/spf13/pflag v1.0.5 // indirect
-	github.com/xdg-go/pbkdf2 v1.0.0 // indirect
-	github.com/xdg-go/scram v1.1.2 // indirect
-	github.com/xdg-go/stringprep v1.0.4 // indirect
-	golang.org/x/crypto v0.11.0 // indirect
-	golang.org/x/exp v0.0.0-20230224173230-c95f2b4c22f2 // indirect
-	golang.org/x/net v0.13.0 // indirect
-	golang.org/x/oauth2 v0.10.0 // indirect
-	golang.org/x/term v0.10.0 // indirect
-	golang.org/x/text v0.11.0 // indirect
-	golang.org/x/time v0.3.0 // indirect
+	github.com/vishvananda/netns v0.0.0-20191106174202-0a2b9b5464df // indirect
+	github.com/xdg/scram v1.0.5 // indirect
+	github.com/xdg/stringprep v1.0.3 // indirect
+	golang.org/x/crypto v0.5.0 // indirect
+	golang.org/x/exp v0.0.0-20230321023759-10a507213a29
+	golang.org/x/net v0.7.0 // indirect
+	golang.org/x/oauth2 v0.0.0-20220411215720-9780585627b5 // indirect
+	golang.org/x/term v0.5.0 // indirect
+	golang.org/x/text v0.7.0 // indirect
+	golang.org/x/time v0.0.0-20220210224613-90d013bbcef8 // indirect
 	google.golang.org/appengine v1.6.7 // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20230711160842-782d3b101e98 // indirect
 	gopkg.in/inf.v0 v0.9.1 // indirect
--- a/go.mod
+++ b/go.mod
@@ -69,16 +69,16 @@ require (
 	github.com/prometheus/common v0.32.1 // indirect
 	github.com/prometheus/procfs v0.7.3 // indirect
 	github.com/spf13/pflag v1.0.5 // indirect
-	github.com/vishvananda/netns v0.0.0-20191106174202-0a2b9b5464df // indirect
-	github.com/xdg/scram v1.0.5 // indirect
-	github.com/xdg/stringprep v1.0.3 // indirect
-	golang.org/x/crypto v0.5.0 // indirect
-	golang.org/x/exp v0.0.0-20230321023759-10a507213a29
-	golang.org/x/net v0.7.0 // indirect
-	golang.org/x/oauth2 v0.0.0-20220411215720-9780585627b5 // indirect
-	golang.org/x/term v0.5.0 // indirect
-	golang.org/x/text v0.7.0 // indirect
-	golang.org/x/time v0.0.0-20220210224613-90d013bbcef8 // indirect
+	github.com/xdg-go/pbkdf2 v1.0.0 // indirect
+	github.com/xdg-go/scram v1.1.2 // indirect
+	github.com/xdg-go/stringprep v1.0.4 // indirect
+	golang.org/x/crypto v0.11.0 // indirect
+	golang.org/x/exp v0.0.0-20230321023759-10a507213a29 // indirect
+	golang.org/x/net v0.13.0 // indirect
+	golang.org/x/oauth2 v0.10.0 // indirect
+	golang.org/x/term v0.10.0 // indirect
+	golang.org/x/text v0.11.0 // indirect
+	golang.org/x/time v0.3.0 // indirect
 	google.golang.org/appengine v1.6.7 // indirect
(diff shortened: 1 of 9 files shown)


joewilliams:
👋 @masonj188 and I are from [ngrok](https://ngrok.com/) and we are starting to use netobserv to get better visibility into our edge traffic. Let us know if this is the sort of patch you are interested in accepting and we'll push it forward. In the future we'll continue to upstream changes we need to netobserv as well.

msherif1234:
> wave @masonj188 and I are from [ngrok](https://ngrok.com/) and we are starting to use netobserv to get better visibility into our edge traffic. Let us know if this is the sort of patch you are interested in accepting and we'll push it forward. In the future we'll continue to upstream changes we need to netobserv as well.

Thanks @joewilliams and @masonj188 for your contribution I will definitely go over your changes, for now I just wanted to make sure u are considering single and dual stack clusters support and being able to handle both address-families IPs

masonj188:
Hi @msherif1234, really appreciate the quick reply!

> just wanted to make sure u are considering single and dual stack clusters support and being able to handle both address-families IPs

Yep! I pushed up some test cases that cover IPv6 addresses as well, should have included those earlier. The list of IPs given in the environment variable are parsed into [netip.Prefix](https://pkg.go.dev/net/netip#Prefix) which should cover both IPv4 and IPv6 addresses.

pkg/agent/filter.go
```diff
@@ -16,10 +60,10 @@ type interfaceFilter struct {
 // initInterfaceFilter allows filtering network interfaces that are accepted/excluded by the user,
```
msherif1234, line 60: pls update the function name in the comment as well

pkg/agent/filter.go
```diff
@@ -2,11 +2,55 @@ package agent
...
+	// the interface from net.InterfaceByName and then calling
+	// .Addrs() on the interface
+	ipsFromIface func(ifaceName string) ([]netip.Prefix, error)
+}
+
+func initIPInterfaceFilter(ips []string, ipsFromIface func(ifaceName string) ([]netip.Prefix, error)) (ipInterfaceFilter, error) {
+	ipIfaceFilter := ipInterfaceFilter{}
+	ipIfaceFilter.ipsFromIface = ipsFromIface
+
+	for _, ip := range ips {
```
msherif1234, line 28: do we need special handling for loopback IPs or LinkLocal ?

eranra:
@joewilliams and @masonj188, I took a quick look at the code, and this will indeed ease configuration, I really like that. Are you talking about the need to configure specific IPs or about configuration by subnets/netmasks? We are talking about interfaces for specific subnets, right? 

Also, can you add needed changes to https://github.com/netobserv/netobserv-ebpf-agent/blob/main/docs/config.md to describe the updated capabilities 

THX for the contribution :-)

pkg/agent/filter.go
```diff
@@ -2,11 +2,55 @@ package agent
...
+		if err != nil {
+			return ipInterfaceFilter{}, fmt.Errorf("error parsing given ip: %s: %w", ip, err)
+		}
+		ipIfaceFilter.allowedIPs = append(ipIfaceFilter.allowedIPs, prefix)
+	}
+
+	return ipIfaceFilter, nil
+}
+
+func (f *ipInterfaceFilter) Allowed(iface string) (bool, error) {
```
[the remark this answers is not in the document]
masonj188: If I understand your question correctly, yep it should work. With these changes a filter just has to implement "Allowed" and will always receive an interface name from the interface monitor when an interface is found/added.

The ipInterfaceFilter receives the interface name, queries for which IPs the interface is associated with, and then checks if any of the IPs were the ones passed in through the "INTERFACE_IPS" env variable, if so it returns true (interface allowed), if not then it returns false.

pkg/agent/filter.go
masonj188: I don't think so, the code uses [net.InterfaceByName()](https://pkg.go.dev/net#InterfaceByName) and then calls [Interface.Addrs()](https://pkg.go.dev/net#Interface.Addrs) which is happy to return loopback or linklocal addresses.

Just running some code that loops through interfaces and calls `.Addrs()` on my machine:
```
if: lo
addrs: 127.0.0.1/8

if: lo
addrs: ::1/128

if: enp6s0
addrs: 10.0.0.100/24

if: enp6s0
addrs: fe80::b981:bd8e:91f6:b3e5/64
```

pkg/agent/agent.go
```diff
@@ -137,10 +138,42 @@ func flowsAgent(cfg *Config,
 	exporter node.TerminalFunc[[]*flow.Record],
 	agentIP net.IP,
 ) (*Flows, error) {
-	// configure allow/deny interfaces filter
-	filter, err := initInterfaceFilter(cfg.Interfaces, cfg.ExcludeInterfaces)
-	if err != nil {
-		return nil, fmt.Errorf("configuring interface filters: %w", err)
+	var filter InterfaceFilter
+
+	if len(cfg.InterfaceIPs) > 0 {
```
msherif1234, line 143: do we need to flag an error if user supplied both `cfg.InterfaceIPs` and `cfg.Interfaces` ?

msherif1234:
@masonj5n I would like to get this PR merged can u pls rebase it Thanks!!

masonj5n:
> @masonj5n I would like to get this PR merged can u pls rebase it Thanks!!

Hi @msherif1234, thanks for the poke! I went ahead and made some changes so it matches my last comment - where you supply an IP CIDR/range and then any interface with an IP in those ranges will make that interface included. I updated some docs and tests to reflect that.

Also, it looks like a recent dependabot update bumped `vmware/go-ipfix` to a version that requires Go 1.21, so my `go mod tidy` bumped up the Go toolchain version to `1.21`.
```
❯ go mod tidy -go=1.19
go: github.com/vmware/go-ipfix@v0.7.0 requires go@1.21, but 1.19 is requested
```

pkg/agent/agent.go
```diff
@@ -415,7 +432,11 @@ func (f *Flows) buildAndStartPipeline(ctx context.Context) (*node.Terminal[[]*fl
 
 func (f *Flows) onInterfaceAdded(iface ifaces.Interface) {
 	// ignore interfaces that do not match the user configuration acceptance/exclusion lists
-	if !f.filter.Allowed(iface.Name) {
+	allowed, err := f.filter.Allowed(iface.Name)
+	if err != nil {
+		alog.WithField("interface", iface).Errorf("encountered error determining if interface is allowed: %v", err)
```
msherif1234, line 437: we should return here , right ?

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/netobserv/netobserv-ebpf-agent
 
-go 1.19
+go 1.21
+
+toolchain go1.21.1
```
msherif1234, line 5: why u need this ? it seems to fail lint check in ur PR pls run
`make lint` on ur view

pkg/agent/agent.go
masonj5n: Yep, thanks for catching that!

go.mod
masonj5n: I didn't realize this but the behavior changed in go 1.21 so that it's a compile error if a dependency has a `go` directive with a higher `go` version than what you're compiling it with:
From https://go.dev/doc/modules/gomod-ref#go:
> The go directive sets the minimum version of Go required to use this module. Before Go 1.21, the directive was advisory only; now it is a mandatory requirement: Go toolchains refuse to use modules declaring newer Go versions.

So, I think things will fail to build on go 1.21 until the `go.mod` is updated like I show here, but I did switch back to go 1.19 and was able to run `go mod tidy` and `go mod vendor` and run `make build`, so I removed this change.

jotak (approved):

PR #112, opened 2023-04-14
NETOBSERV-974 Add SASL support

jotak:
SASL is a common protocol used for authentication with Kafka, that strimzi supports.
This PR adds support to SASL configuration when the agent is used with Kafka. To enable it, set `KAFKA_ENABLE_SASL` to true. The path to clientID (`KAFKA_SASL_CLIENT_ID_PATH`) and to a Secret (`KAFKA_SASL_CLIENT_SECRET_PATH`) must be configured. `KAFKA_SASL_TYPE` can be either "plain" or "scramSHA512".

### DIFF
--- a/pkg/agent/config.go
+++ b/pkg/agent/config.go
@@ -126,6 +126,14 @@ type Config struct {
 	KafkaTLSUserCertPath string `env:"KAFKA_TLS_USER_CERT_PATH"`
 	// KafkaTLSUserKeyPath is the path to the user (client) private key for mTLS connections
 	KafkaTLSUserKeyPath string `env:"KAFKA_TLS_USER_KEY_PATH"`
+	// KafkaEnableSASL set true to enable SASL auth
+	KafkaEnableSASL bool `env:"KAFKA_ENABLE_SASL" envDefault:"false"`
+	// KafkaSASLType type of SASL mechanism: plain or scramSHA512
+	KafkaSASLType string `env:"KAFKA_SASL_TYPE" envDefault:"plain"`
+	// KafkaSASLClientIDPath is the path to the client ID (username) for SASL auth
+	KafkaSASLClientIDPath string `env:"KAFKA_SASL_CLIENT_ID_PATH"`
+	// KafkaSASLClientSecretPath is the path to the client secret (password) for SASL auth
+	KafkaSASLClientSecretPath string `env:"KAFKA_SASL_CLIENT_SECRET_PATH"`
 	// ProfilePort sets the listening port for Go's Pprof tool. If it is not set, profile is disabled
 	ProfilePort int `env:"PROFILE_PORT"`
 }
--- a/pkg/agent/sasl.go
+++ b/pkg/agent/sasl.go
@@ -30,7 +30,7 @@ func buildSASLConfig(cfg *Config) (sasl.Mechanism, error) {
 	case "scramSHA512":
 		mechanism, err = scram.Mechanism(scram.SHA512, strID, strPwd)
 	default:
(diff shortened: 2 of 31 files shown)


jotak:
linter fails due to cyclo: taking this opportunity to ask you, @msherif1234 @praveingk what's your thoughts on this. Personally I hate this linter metric, I think a long and dumb function is better than 10 small and dumb functions. Especially when it doesn't make code more reused. So if it was just me I would just raise the golangci-lint cyclop max value to make linter pass. But as you're more actively contributing to this repo, I'll follow what you can suggest :-)
(I can also defintely create new dumb functions to make linter happier)

pkg/agent/config.go
```diff
@@ -126,6 +126,14 @@ type Config struct {
...
 	// KafkaTLSUserKeyPath is the path to the user (client) private key for mTLS connections
 	KafkaTLSUserKeyPath string `env:"KAFKA_TLS_USER_KEY_PATH"`
+	// KafkaEnableSASL set true to enable SASL auth
+	KafkaEnableSASL bool `env:"KAFKA_ENABLE_SASL" envDefault:"false"`
+	// KafkaSASLType type of SASL mechanism: plain or scramSHA512
+	KafkaSASLType string `env:"KAFKA_SASL_TYPE" envDefault:"plain"`
+	// KafkaSASLClientIDPath is the path to the client ID (username) for SASL auth
+	KafkaSASLClientIDPath string `env:"KAFKA_SASL_CLIENT_ID_PATH"`
+	// KafkaSASLClientSecretPath is the path to the client secret (password) for SASL auth
+	KafkaSASLClientSecretPath string `env:"KAFKA_SASL_CLIENT_SECRET_PATH"`
```
msherif1234, line 136: is there a default value ?

pkg/agent/sasl.go
```diff
@@ -0,0 +1,39 @@
...
+	var mechanism sasl.Mechanism
+	switch cfg.KafkaSASLType {
+	case "plain":
+		mechanism = plain.Mechanism{Username: strID, Password: strPwd}
+	case "scramSHA512":
+		mechanism, err = scram.Mechanism(scram.SHA512, strID, strPwd)
+	default:
+		return nil, fmt.Errorf("unknown SASL type: %s", cfg.KafkaSASLType)
+	}
+	if err != nil {
```
msherif1234, line 35: there is no need for this block since u already return on err

msherif1234:
> linter fails due to cyclo: taking this opportunity to ask you, @msherif1234 @praveingk what's your thoughts on this. Personally I hate this linter metric, I think a long and dumb function is better and more readable than 10 small and dumb functions. Especially when it doesn't make code more reused. So if it was just me I would just raise the golangci-lint cyclop max value to make linter pass. But as you're more actively contributing to this repo, I'll follow what you can suggest :-) (I can also defintely create new dumb functions to make linter happier)

I would just creating 3 functions one for grpc , kafka and one for ipfix that take protocol as arg, in general even though the function is plain straight fwd but its not good practice having lengthy functions IMO but again this more of nit preference I will go with whatever the rest agree with

pkg/agent/sasl.go
jotak: actually, the line above was setting an error; I modified this part so it's more visible that an error might be produced

msherif1234:
pls add more context to the PR description so we can know what this PR did for reference Thanks!!

jotak:
Thanks @msherif1234 

/approve

jotak:
Just to clarify, I set no-doc/QE labels as it has no impact downstream unless it's configured from the operator - there's an epic for that which isn't planned yet.

PR #113, opened 2023-04-19
NETOBSERV-1255 Adding Raw Packet export capability as Packet Capture Agent(PCA)

shach33:
This PR extends the agent to export raw packets through the network.  We call it Packet Capture Agent(PCA). 

This functionality is turned off by default. I have added 2 new environments to enable this. 
`set ENABLE_PCA=true` to activate packet cpature. 
set PCA_FILTER=[ipproto,port number] to indicate the filter for packet capture,  e.g.` export PCA_FILTER=tcp,22`

The PCA_FILTER defaults to udp DNS(udp,53). If ENABLE_PCA is true, but filters are not specified, by default it captures DNS packets. 

The captured packets are written to a socket in PCAP format.


eranra:
@msherif1234, for awareness --- the idea behind this PR is to become the basis for both PANO (of course) but also for other Packet level network observability requirements, such as "selective" packet mirroring - functionality such as network tracing etc. So generalization and customization of the content are key ( hope this make sense to you )

bpf/flows.c
```diff
@@ -69,6 +69,17 @@ struct {
...
 } direct_flows SEC(".maps");
 
+
+// Perf Buffer to submit packet payloads to userspace
+struct {
+    
+    __uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
+	__uint(key_size, sizeof(__u32));
+	__uint(value_size, sizeof(__u32));
+    __uint(max_entries, 80);
```
msherif1234, line 79: just 80 entries ? is this experimental value ?

bpf/flows.c
```diff
@@ -325,6 +336,66 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
...
+    if ((void *)ip + sizeof(*ip) > data_end) {
+       return TC_ACT_UNSPEC;	
+    }
+
+    udp_data = (void *)ip + sizeof(*ip);
+    if ((void *)udp_data + sizeof(*udp_data) > data_end) {
+       return TC_ACT_UNSPEC;	
+    }
+
+    if (eth->h_proto != __bpf_htons(ETH_P_IP)) {
```
msherif1234, line 364: what about IPv6 ?

bpf/flows.c
```diff
@@ -325,6 +336,66 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
...
+    if ((void *)eth + sizeof(*eth) > data_end) {
+       return TC_ACT_UNSPEC;	
+    }
+
+    ip = data + sizeof(*eth);
+    if ((void *)ip + sizeof(*ip) > data_end) {
+       return TC_ACT_UNSPEC;	
+    }
+
+    udp_data = (void *)ip + sizeof(*ip);
```
msherif1234, line 359: only udp ?

bpf/flows.c
```diff
@@ -325,6 +336,66 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
...
+    }
+
+    //Only analyze UDP packets
+    if (ip->protocol != IPPROTO_UDP ) {
+       return TC_ACT_UNSPEC;	
+    }
+
+    __be16 port = udp_data->dest;
+    //TODO: Update port number/filters to be read from ENV variable
+    __be16 portFromPanoFilter = __bpf_htons(53);
```
msherif1234, line 375: is the oplan to use only one port as filter or we can also provide range ?

bpf/flows.c
```diff
@@ -325,6 +336,66 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
...
+    bpf_printk("Into export packet payload %d (packet) -- %d\n", port , portFromPanoFilter);
+       
+    if (port == portFromPanoFilter) {
+       meta.if_index = skb->ifindex;
+       meta.pkt_len = data_end - data;
```
msherif1234, line 380: u can get the length directly from sbk `skb->len`

bpf/flows.c
```diff
@@ -325,6 +336,66 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
...
+       meta.payload = data;
```
msherif1234, line 381: I don't think its a good idea to share the pkt kernel pointer with userspace
instead u can pass in flags to event_out with the header size u wanted
as u already doing and kernel will copy the header just after ur metadata struct

bpf/flows.c
```diff
@@ -325,6 +336,66 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
...
+SEC("tc_pano_ingress")
+int ingress_pano_parse (struct __sk_buff *skb) {
+    return export_packet_payload(skb);
+}
+
+SEC("tc_pano_egress")
+int egress_pano_parse (struct __sk_buff *skb) {
+    return export_packet_payload(skb);
+}
+
```
msherif1234, line 398: we were discussing plans on how to extend the kernel hook 
so the direction was to create folder / the different hook
so we can have some modularity I kinda of started some of this in #106

pkg/agent/ip.go
```diff
@@ -113,7 +113,7 @@ func findAddress(addrs []net.Addr, ipType string) (net.IP, bool) {
 }
 
 func getIP(pip net.IP, ipType string) (net.IP, bool) {
-	if pip == nil || pip.IsLoopback() || pip.IsLinkLocalUnicast() {
+	if pip == nil || pip.IsLoopback() {
```
msherif1234, line 116: why did u remove LL check ? this will break ipv6 cluster only on baremetal ?

msherif1234 (commented):
I took initial look
I see ur created a parallel agent instead of extending the current one after u check the new config knob can u explain the reasoning ?

bpf/flows.c
shach33: Eventually we would want this filter to be able to serve a number of possible subsets of packets. 
Ad of now, we only want to use it for port 53 for DNS.

bpf/flows.c
shach33: For current use case, with DNS over UDP.

bpf/flows.c
shach33: This is an experimental value. Reverting.

bpf/flows.c
shach33: Will make the change.

pkg/agent/ip.go
shach33: This was not an intentional change. Reverting.

shach33:
> I took initial look I see ur created a parallel agent instead of extending the current one after u check the new config knob can u explain the reasoning ?

The current agent works at the level of flows and flow metrics. With PANO, we wish to use a 'packet' as the unit of communication. The struct elements of flow metrics are no longer required. Hope this helps.

msherif1234:
> > I took initial look I see ur created a parallel agent instead of extending the current one after u check the new config knob can u explain the reasoning ?
> 
> The current agent works at the level of flows and flow metrics. With PANO, we wish to use a 'packet' as the unit of communication. The struct elements of flow metrics are no longer required. Hope this helps.

yeah I am aware of that as u know I went through the same trouble when I prototyped TCPlife span hook but was able to refactor the current agent so it can handle both probably something we need to poll the team on @jotak @eranra @praveingk wdyt ? will be great if we somehow abstract support to perf map as we have other use cases as well that wanted to use it like the one I mentioned above and more will come and will be nice to share this so all hooks need to generate perf event will reuse

bpf/flows.c
shach33: Would it be possible to explain this a bit more. I am not sure I understand the alternative you are suggesting.

bpf/flows.c
shach33: So, this current PR is essentially part of PANO being built on top of the agent.  This is also the reason to write an entire parallel pipeline for this. We can connect over a call if that helps.

bpf/flows.c
dushyantbehl: Can you make this a macro, hardcode of the port inside the code is less readable.
On that front if comparing ports then enter reverse endian format so that you save on on `bpf_htons` while comparing.

Also might be nitpicking but can you rename the variable from `portFromPanoFilter` to `filter_port` or something else, C naming convention does not like camel case.

bpf/flows.c
dushyantbehl: I also agree that we should perfrom hook point separation in folders, but for this particular PR since the program sections are defined separately they should not get loaded along with our previous hook points.

bpf/flows.c
```diff
@@ -325,6 +335,64 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
...
+    __be16 dport = udp_data->dest;
+    __be16 sport = udp_data->source;
+    //TODO: Update port number/filters to be read from ENV variable
+    __be16 portFromPanoFilter = __bpf_htons(53);
+    
+    if (dport == portFromPanoFilter || sport == portFromPanoFilter) {
+        meta.if_index = skb->ifindex;
+        meta.pkt_len = data_end - data;
+        meta.payload = data;
+        bpf_perf_event_output(skb, &packet_payloads, ((u64) meta.pkt_len << 32) | BPF_F_CURRENT_CPU, &meta, sizeof(meta));
```
dushyantbehl, line 379: @shach33 So currently you are just sending the packet data pointer to userspace, imo this has a race condition here, if some piece of code calls skb_free over the packet in between ebpf sending it and userspace reading it there is a possibility of null pointer exception (or segmentation fault) in the userspace agent. 
If my understanding is correct can you explain the reasoning for sending the pointer and not deepcopy the packet.

cc @msherif1234 do you think this assertion that there is a race condition here is correct?

cmd/netobserv-ebpf-agent.go
```diff
@@ -39,22 +39,46 @@ func main() {
 
 	logrus.WithField("configuration", fmt.Sprintf("%#v", config)).Debugf("configuration loaded")
 
-	flowsAgent, err := agent.FlowsAgent(&config)
-	if err != nil {
-		logrus.WithError(err).Fatal("can't instantiate NetObserv eBPF Agent")
-	}
+	if config.EnablePano {
```
dushyantbehl, line 42: Shouldn't the check be which agent to enable rather than enablePano?

Is there a scenario that we might need both the agents enabled from the userspace?

cmd/netobserv-ebpf-agent.go
```diff
@@ -39,22 +39,46 @@ func main() {
...
+		packetsAgent, err := agent.PacketsAgent(&config)
+		if err != nil {
```
dushyantbehl, line 44: If this code is common can we move it to a single function so less duplication is present.

pkg/agent/agent.go
```diff
@@ -397,3 +418,236 @@ func (f *Flows) onInterfaceAdded(iface ifaces.Interface) {
...
+
+	plog.Info("[PANO]acquiring Agent IP")
+	agentIP, err := fetchAgentIP(cfg)
+	if err != nil {
+		return nil, fmt.Errorf("acquiring Agent IP: %w", err)
+	}
+
+	// configure selected exporter
+	packetexportFunc, err := buildPacketExporter(cfg)
+	// Log statements to be removed eventually
```
dushyantbehl, line 459: Which log statements do you mean?
Can you remove them now if the work is done. else please remove these comments.
Like one below as well.

pkg/agent/agent.go
```diff
@@ -397,3 +418,236 @@ func (f *Flows) onInterfaceAdded(iface ifaces.Interface) {
...
+		}
+		return iface
+	}
+
+	perfTracer := flow.NewPerfTracer(fetcher, cfg.CacheActiveTimeout)
+
+	packetbuffer := flow.NewPerfBuffer(cfg.CacheMaxFlows, cfg.CacheActiveTimeout)
+
+	return &Packets{
+		// TODO : flows -> Packets
```
dushyantbehl, line 514: remove?

pkg/agent/agent.go
```diff
@@ -397,3 +418,236 @@ func (f *Flows) onInterfaceAdded(iface ifaces.Interface) {
...
+			Debug("[PANO]interface does not match the allow/exclusion filters. Ignoring")
+		return
+	}
+	plog.WithField("[PANO]interface", iface).Info("interface detected. Registering packets ebpfFetcher")
+	if err := p.ebpf.Register(iface); err != nil {
+		plog.WithField("[PANO]interface", iface).WithError(err).
+			Warn("can't register packet ebpfFetcher. Ignoring")
+		return
+	}
+}
```
dushyantbehl, line 653: Overall I feel like there is high level of duplication between flows agent and pano agent so it makes sense to remove it and instantiate through a common agents interface. Else we will have big overhead of code maintenance.

pkg/agent/agent_test.go
```diff
@@ -226,7 +226,8 @@ func testAgent(t *testing.T, cfg *Config) *test.ExporterFake {
 	key2Metrics := []ebpf.BpfFlowMetrics{
 		{Packets: 7, Bytes: 33, StartMonoTimeTs: now, EndMonoTimeTs: now + 2_000_000_000},
 	}
-
+	_ = key1Metrics
+	_ = key2Metrics
```
dushyantbehl, line 230: Why do we need this line?

pkg/flow/packet_record.go
```diff
@@ -0,0 +1,40 @@
...
+	len uint32,
+) *PacketRecord {
+	pr := PacketRecord{}
+	pr.Stream = make([]byte, len)
+	pr.Stream = stream
+	return &pr
+}
+
+// ReadPacketFrom reads a PacketRecord from a binary source, in LittleEndian order
+func ReadRawPacketFrom(reader io.Reader) (*PacketRecord, error) {
```
dushyantbehl, line 28: Can you rename this to what it does `ReadRawPacketFromBuffer` or just `ReadRawPacket`

dushyantbehl (changes requested):
@shach33 can you also remove the `temp1.pcap` file which I assume got pushed along with your commits.

dushyantbehl:
> 

I belive yes we can extract out common code and enable one or both functionalities in the agents rather than extending the code which can create code duplication.

bpf/flows.c
```diff
@@ -1,17 +1,17 @@
...
-           Eviction for non-tcp flows need to done by userspace
-        3) When the map is full, we send the new flow entry to userspace via ringbuffer,
-            until an entry is available.
-        4) When hash collision is detected, we send the new entry to userpace via ringbuffer.
+   Flows v2. A Flow-metric generator using TC.
+
+   This program can be hooked on to TC ingress/egress hook to monitor packets
+   to/from an interface.
+
+Logic:
```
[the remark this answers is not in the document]
shach33: Reverting.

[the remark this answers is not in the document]
shach33: Reverted.

shach33:
Performed first level refactoring.

pkg/agent/agent.go
dushyantbehl: @shach33 is this comment to be fixed later? or was it missed.

bpf/flows.c
```diff
@@ -69,6 +73,16 @@ struct {
...
 
+
+// Perf Buffer to submit packet payloads to userspace
+struct {
+    
+    __uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
+	__uint(key_size, sizeof(__u32));
+	__uint(value_size, sizeof(__u32));
+} packet_record   SEC(".maps");
+
```
dushyantbehl, line 84: Indent fix please

pkg/agent/agent.go
```diff
@@ -397,3 +425,188 @@ func (f *Flows) onInterfaceAdded(iface ifaces.Interface) {
...
+
+	return pcapStreamer.ExportFlows, err
+
+}
+
+// Run a Packets agent. The function will keep running in the same thread
+// until the passed context is canceled
+func (p *Packets) Run(ctx context.Context) error {
+	p.status = StatusStarting
+	plog.Info("[NEW] starting Packets agent")
```
dushyantbehl, line 526: @shach33 did you intend to keep this `[NEW]` or this is a debug message?

bpf/flows.c
eranra: Fixed

dushyantbehl:
This branch needs a code review. Waiting for @shach33 to commit final version before providing the review

.gitignore
```diff
@@ -4,3 +4,5 @@
 bin/
 e2e-logs
 ebpf-agent.tar
+*.o
```
jotak, line 7: this is unclear to me why we would ignore `.o` files, I think we shouldn't?

cmd/netobserv-ebpf-agent.go
```diff
@@ -39,22 +53,30 @@ func main() {
 
 	logrus.WithField("configuration", fmt.Sprintf("%#v", config)).Debugf("configuration loaded")
 
-	flowsAgent, err := agent.FlowsAgent(&config)
-	if err != nil {
-		logrus.WithError(err).Fatal("can't instantiate NetObserv eBPF Agent")
-	}
+	if config.EnablePCA {
```
jotak, line 56: so this means the agent is started either for flow logs or for PCA, but cannot do both at the same time, right?
Do you see a technical limitation for doing so, or is that just a choice and we could change that any time?

pkg/ebpf/tracer.go
```diff
@@ -127,6 +137,7 @@ func NewFlowFetcher(cfg *FlowFetcherConfig) (*FlowFetcher, error) {
 		var newObjects NewBpfObjects
 		// remove tcpdrop hook from the spec
 		delete(spec.Programs, tcpDropHook)
+		delete(spec.Programs, )
```
jotak, line 140: there are some compilation errors here and above, I guess it wasn't intended to be commited?

pkg/ebpf/tracer.go
```diff
@@ -434,3 +472,181 @@ func (m *FlowFetcher) LookupAndDeleteMap() map[BpfFlowId]*BpfFlowMetrics {
...
+
+	pcaPort := 0
+	pcaProto := 0
+	filters := strings.Split(pcaFilters, ",")
+	if filters[0] == "tcp" {
+		pcaProto = syscall.IPPROTO_TCP
+	} else {
+		pcaProto = syscall.IPPROTO_UDP
+	}
+	pcaPort, _ = strconv.Atoi(filters[1])
```
jotak, line 512: the error should be caught & reported if the configuration is incorrect

pkg/ebpf/tracer.go
```diff
@@ -434,3 +472,181 @@ func (m *FlowFetcher) LookupAndDeleteMap() map[BpfFlowId]*BpfFlowMetrics {
...
+
+	if err := spec.RewriteConstants(map[string]interface{}{
+		constPcaPort:  uint32(pcaPort),
+		constPcaProto: uint8(pcaProto),
+	}); err != nil {
+		return nil, fmt.Errorf("rewriting BPF constants definition: %w", err)
+	}
+	plog.Debugf("PCA Filters: %d, %d Incoming: %s", pcaProto, pcaPort, pcaFilters)
```
jotak, line 520: Maybe raise as Info log, as it's a one-time log that helps to retrieve the configuration applied

pkg/agent/agent.go
```diff
@@ -417,3 +387,248 @@ func (f *Flows) onInterfaceAdded(iface ifaces.Interface) {
...
+		exporter:       packetexporter,
+	}, nil
+}
+
+func buildPacketExporter(cfg *Config) (node.TerminalFunc[[]*flow.PacketRecord], error) {
+	if cfg.TargetPort == 0 {
+		return nil, fmt.Errorf("missing target port: %d",
+			cfg.TargetPort)
+	}
+	pcapStreamer, err := exporter.StartPCAPSend(fmt.Sprintf("%d", cfg.TargetPort))
```
jotak, line 535: I feel it is misleading to reuse the `cfg.TargetPort` here as this is used for the agent _listening_ as a server, unlike its usage for flows where it is use for sending (as a client). I guess that's why I was misguided the other day and I was trying to configure something server-side to receive packets.

I would suggest either to create new parameters to clarify that the agent will be _listening_ on a port, or to reverse the approach and have the agent send the pcap to a receiver. @shach33  @msherif1234 @dushyantbehl wdyt is better?

shach33:
Added comments. Test

jotak:
FYI, I've created this gist that showcases how that could work in a kube context: https://gist.github.com/jotak/2a4a5374ee0ab780aeb31bf6ee9a23de
Deployed as a new agent pod (so, independent from the flow collection - which in a sense is better as it offers separation of concerns). By the way, there was no conflicts with the flow-collector agents. The gist illustrates a way to work with the terminal (a next step would be how to integrate that with the console.. let's keep that for later :-) )

pkg/exporter/packets.go
```diff
@@ -0,0 +1,136 @@
...
+const versionMajor = 2
+const versionMinor = 4
+const nanosPerMicro = 1000
+
+var plog = logrus.WithField("component", "packet/Packets")
+
+// Setting Snapshot length to 0 sets it to maximum packet size
+var snapshotlen int32
+
+func WriteFileHeader(snaplen uint32, linktype layers.LinkType, conn net.Conn) error {
```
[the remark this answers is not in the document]
dushyantbehl: if this is internal function you can start with smallcase.

pkg/exporter/packets.go
```diff
@@ -0,0 +1,136 @@
...
+		plog.Fatal(err)
+	}
+	plog.Debugf("Writing File Header")
+	for packetRecord := range in {
+		if len(packetRecord) > 0 {
+			for _, packet := range packetRecord {
+				packetStream := packet.Stream
+				packetTimestamp := packet.Time
+				if len(packetStream) != 0 {
+					plog.Debugf("TS from packet: %s", packetTimestamp)
```
dushyantbehl, line 121: please remove this log message.

pkg/flow/packet_record.go
```diff
@@ -0,0 +1,56 @@
...
+func ReadRawPacket(reader io.Reader) (*PacketRecord, error) {
+	var pr PacketRecord
+	currentTime := time.Now()
+	monotonicTimeNow := monotime.Now()
+	getLen := make([]byte, 2)
+	packetTimestamp := make([]byte, 8)
+	// Read IfIndex and discard it: To be used in other usecases
+	_ = binary.Read(reader, binary.LittleEndian, make([]byte, 2))
+	// Read Length of packet
+	_ = binary.Read(reader, binary.LittleEndian, getLen)
```
dushyantbehl, line 44: packet length should be 32 bits.

pkg/flow/packet_record.go
```diff
@@ -0,0 +1,56 @@
...
+
+// ReadRawPacket reads a PacketRecord from a binary source, in LittleEndian order
+func ReadRawPacket(reader io.Reader) (*PacketRecord, error) {
+	var pr PacketRecord
+	currentTime := time.Now()
+	monotonicTimeNow := monotime.Now()
+	getLen := make([]byte, 2)
+	packetTimestamp := make([]byte, 8)
+	// Read IfIndex and discard it: To be used in other usecases
+	_ = binary.Read(reader, binary.LittleEndian, make([]byte, 2))
```
dushyantbehl, line 42: if index should be 32 bits

pkg/flow/packet_record.go
```diff
@@ -0,0 +1,56 @@
...
+	// Read Length of packet
+	_ = binary.Read(reader, binary.LittleEndian, getLen)
+	plog.Debugf("Reading packet of length: %d", binary.LittleEndian.Uint16(getLen))
```
dushyantbehl, line 45: delete

pkg/flow/tracer_perf.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		return fmt.Errorf("reading from perf event array: %w", err)
+	}
+	// Parses the perf event entry into an Event structure.
+	readFlow, err := ReadRawPacket(bytes.NewBuffer(event.RawSample))
+	if err != nil {
+		return fmt.Errorf("parsing data received from the perf event array: %w", err)
+	}
+	if debugging {
+		m.stats.logRingBufferFlows(false)
```
dushyantbehl, line 72: Why are we calling `logRingBufferFlows` for perf array?
Are they doing the same functionality?

dushyantbehl (changes requested):
This PR needs changes. Please find my review comments and message me if any clarification is needed.

bpf/pca.h
```diff
@@ -0,0 +1,96 @@
...
+static int attach_packet_payload(void *data, void *data_end, struct __sk_buff *skb, u64 current_time ){
+        payload_meta meta;
+        u64 flags = BPF_F_CURRENT_CPU;
+        // Enable the flag to add packet header
+        // Packet payload follows immediately after the meta struct
+        u32 packetSize = (__u32)(data_end-data);
+        // For packets which are allocated non-linearly struct __sk_buff does not necessarily 
+        // has all data lined up in memory but instead can be part of scatter gather lists. 
+        // This command pulls data from the buffer but incurs data copying penalty.
+        if (packetSize < skb->len){
```
dushyantbehl, line 16: Shouldn't the check be `<=`?

bpf/pca.h
```diff
@@ -0,0 +1,96 @@
...
+            packetSize = skb->len;
+            bpf_skb_pull_data(skb, skb->len);
```
dushyantbehl, line 18: @shach33 can you add a check here if this return any error or not, it return 0 on success and else error.

bpf/pca.h
```diff
@@ -0,0 +1,96 @@
...
+        if ((void *)tcp_header + sizeof(*tcp_header) > data_end) {
+            return TC_ACT_UNSPEC;
+        }
+        if (tcp_header->source == bpf_htons(pca_port) || tcp_header->dest == bpf_htons(pca_port)){
+            return attach_packet_payload(data, data_end, skb, current_time);
+        }
+        return TC_ACT_OK;
+    }
+
+    udp_header = (void *)ip + sizeof(*ip);
```
dushyantbehl, line 75: @shach33
This just doesn't seem right to me, the code is still duplicating something which is very small. 
Can you create a function which takes in `ip->protocol` and the pointer to `(void *)ip + sizeof(*ip)` and internally does an if-then-else loop on the `ip->protocol` and checks if source or dest port match pca port. 
If so return true if need to attach packet payload else false.

Will be much more readable and won't have duplication.

shach33: Done.

shach33:
Performed Resource analysis on the PR with 6 different use cases. 
1. base netobserv:main
2. PR with PCA disabled
3. PR with PCA enabled - invalid filter 
4. PR with PCA enabled - filter tcp,80 
5. PR with PCA enabled - filter udp,53
6. PR with PCA enabled - filter tcp,443 

Images are annotated with the indices of testcases. 

CPU usage
Memory Usage

bpf/pano.h
```diff
@@ -0,0 +1,64 @@
+#ifndef __PANO_H__
```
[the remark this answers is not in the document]
shach33: Updated to PCA.

shach33:
> 

Removed.

bpf/pano.h
```diff
@@ -0,0 +1,64 @@
...
+    void *data = (void *)(long)skb->data;
+    payload_meta meta;
+    struct ethhdr *eth  = data;
+    struct iphdr  *ip;
+    struct udphdr *udp_data;
+    __u64 flags = BPF_F_CURRENT_CPU;
+    __u16 headerSize;
+    __u64 packet_len;
+
+
```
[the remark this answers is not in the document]
shach33: Fixed.

pkg/ebpf/tracer.go
```diff
@@ -434,3 +475,184 @@ func (m *FlowFetcher) LookupAndDeleteMap() map[BpfFlowId]*BpfFlowMetrics {
...
+	return p.perfReader.Read()
+}
+
+func (p *PacketFetcher) LookupAndDeleteMap() map[int][]*byte {
+	packetMap := p.objects.PacketRecord
+	iterator := packetMap.Iterate()
+	packets := make(map[int][]*byte, p.cacheMaxSize)
+
+	var id int
+	var metrics []*byte
```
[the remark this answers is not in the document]
shach33: Done

cmd/netobserv-ebpf-agent.go
jotak: sounds good

examples/packetcapture-dump/README.md
```diff
@@ -0,0 +1,45 @@
...
+```bash
+make build
+```
+Build the packetcapture-dump-collector (the client that receives full packets from the agent and writes to a pcap file) using:
+```bash
+go build -mod vendor -o bin/packetcapture-client examples/packetcapture-dump/client/packetcapture-client.go  
+```
+Start the agent using:
+```bash
+sudo FLOWS_TARGET_HOST=127.0.0.1 FLOWS_TARGET_PORT=9990 ENABLE_PCA=true PCA_FILTER=tcp,22 ./bin/netobserv-ebpf-agent
```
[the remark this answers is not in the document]
jotak: `PCA_SERVER_PORT` sounds good to me

docs/config.md
```diff
@@ -65,6 +65,10 @@ The following environment variables are available to configure the NetObserv eBF
   If it is not set, profile is disabled.
 * `ENABLE_RTT` (default: `false` disabled). If `true` enables RTT calculations for the captured flows in the ebpf agent.
   See [docs](./rtt_calculations.md) for more details on this feature.
+* `ENABLE_PCA` (default: `false` disabled). If `true` enables Packet Capture Agent. 
+* `PCA_FILTER` (default: `none`). Works only when `ENABLE_PCA` is set. Accepted format <protocol,portnumber>. Example 
+  `PCA_FILTER=tcp,22`.
+* `PCA_SERVER_PORT` (default: 0). Works only when `ENABLE_PCA` is set. Agent opens PCA Server at this port. A collector can connect to it and recieve filtered packets as pcap stream. The filter is set using `PCA_FILTER`.
```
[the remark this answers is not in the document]
jotak: yeah I would prefer "SERVER" here to make it clear that in order to collect pcaps, a client is needed (unlike the netflows that are sent to FLP server)

jotak:
The PR is isolated behind a feature gate, not enabled yet from the operator, hence setting labels no-doc / no-qe

jotak:
> Performed Resource analysis on the PR with 6 different use cases.
> 
>     1. base netobserv:main
> 
>     2. PR with PCA disabled
> 
>     3. PR with PCA enabled - invalid filter
> 
>     4. PR with PCA enabled - filter tcp,80
> 
>     5. PR with PCA enabled - filter udp,53
> 
>     6. PR with PCA enabled - filter tcp,443
> 

@shach33 if I understand correctly your perf measurements, the current ebpf agent (without your PR) consumes twice as much memory than with your PR with PCA disabled? (~ 140MB vs 280MB). This would be good news but is quite unexpected, how can we explain that?

bpf/maps_definition.h
```diff
@@ -31,4 +31,12 @@ struct {
...
 } flow_sequences SEC(".maps");
 
+//PerfEvent Array for Payloads
+struct {
+    __uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
+    __uint(max_entries, MAX_CPUS);
+	__type(key, __u32);
+	__type(value, __u32);
+} packet_record SEC(".maps");
+
```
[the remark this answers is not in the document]
jotak: this is all preallocated? Can't we use here the same `BPF_F_NO_PREALLOC` as for other map types?

shach33:
Kernel does not support `BPF_F_NO_PREALLOC` for LRU type maps.

bpf/maps_definition.h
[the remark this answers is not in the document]
shach33: Kernel does not support BPF_F_NO_PREALLOC for LRU type maps.
https://github.com/cilium/ebpf/issues/1072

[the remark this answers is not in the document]
msherif1234: `BPF_F_NO_PREALLOC` not avaliable for all map types

bpf/pca.h
```diff
@@ -0,0 +1,106 @@
...
+
+    // Record the current time.
+    __u64 current_time = bpf_ktime_get_ns();
+    
+    if ((void *)eth + sizeof(*eth) > data_end) {
+       return TC_ACT_UNSPEC;
+    }
+
+    // Only IPv4 and IPv6 packets captured
+    if (eth->h_proto != bpf_htons(ETH_P_IP) && (eth->h_proto != bpf_htons(ETH_P_IPV6))) {
```
msherif1234, line 80: might be more efficient instruction wise to do
```c
u16 ethType = bpf_ntohs(eth->h_proto);
if (ethType != ETH_P_IP && ethType != ETH_P_IPV6) {
}
```

cmd/netobserv-ebpf-agent.go
```diff
@@ -18,6 +18,20 @@ import (
...
+	ctx, canceler := context.WithCancel(context.Background())
+	// Subscribe to signals for terminating the program.
+	go func() {
+		stopper := make(chan os.Signal, 1)
+		signal.Notify(stopper, os.Interrupt, syscall.SIGTERM)
+		<-stopper
+		canceler()
+	}()
+	return ctx
+}
```
msherif1234, line 33: why did u add this ? for standalone testing ?

pkg/ebpf/tracer.go
```diff
@@ -117,12 +124,19 @@ func NewFlowFetcher(cfg *FlowFetcherConfig) (*FlowFetcher, error) {
 		return nil, fmt.Errorf("rewriting BPF constants definition: %w", err)
 	}
 
+	//Deleting specs for PCA
```
jotak, line 127:
```suggestion
	//Deleting specs for PCA


	// Always set pcaRecordsMap to the minimum in FlowFetcher - PCA and Flow Fetcher are mutually exclusive.
	spec.Maps[pcaRecordsMap].MaxEntries = 1
```

jotak (commented):
@shach33 : here's a suggestion as discussed on slack. This is to minimize the memory allocation when PCA isn't used.

bpf/pca.h
```diff
@@ -0,0 +1,106 @@
...
+        // Enable the flag to add packet header
+        // Packet payload follows immediately after the meta struct
+        u32 packetSize = (__u32)(data_end-data);
+        // For packets which are allocated non-linearly struct __sk_buff does not necessarily 
+        // has all data lined up in memory but instead can be part of scatter gather lists. 
+        // This command pulls data from the buffer but incurs data copying penalty.
+        if (packetSize < skb->len){
+            packetSize = skb->len;
+            if (bpf_skb_pull_data(skb, skb->len)){
+                return TC_ACT_UNSPEC;
```
msherif1234, line 19: I think we come to this case for TCP DNS have u tried to capture DNS over TCP and measured the CPU usage ? u can simply do `dig google.om +tcp` to do DNS over TCP

bpf/maps_definition.h
```diff
@@ -30,6 +30,14 @@ struct {
     __uint(map_flags, BPF_F_NO_PREALLOC);
 } flow_sequences SEC(".maps");
 
+//PerfEvent Array for Packet Payloads
+struct {
+    __uint(type, BPF_MAP_TYPE_PERF_EVENT_ARRAY);
+	__type(key, int);
```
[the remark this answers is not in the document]
shach33: dome

bpf/pca.h
```diff
@@ -0,0 +1,106 @@
...
+        }
+        // Set flag's upper 32 bits with the size of the paylaod and the bpf_perf_event_output will 
+        // attach the specified amount of bytes from packet to the perf event
+        // https://github.com/xdp-project/xdp-tutorial/tree/9b25f0a039179aca1f66cba5492744d9f09662c1/tracing04-xdp-tcpdump       
+        flags |= (__u64)packetSize << 32;
+
+        meta.if_index = skb->ifindex;
+        meta.pkt_len = packetSize;
+        meta.timestamp = current_time;
+        if (bpf_perf_event_output(skb, &packet_record, flags, &meta, sizeof(meta))){
```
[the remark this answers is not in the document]
shach33: this was an oversight. Fixed it.

bpf/pca.h
```diff
@@ -0,0 +1,106 @@
...
+        destPort = tcp_header->dest;
+    }
+    else{
+        struct udphdr *udp_header = ipheaderend;
+        if ((void *)udp_header + sizeof(*udp_header) > data_end) {
+            return false;	
+        }
+        sourcePort = udp_header->source;
+        destPort = udp_header->dest;
+    }
```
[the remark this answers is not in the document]
shach33: Updated. Only checks for tcp/udp, for anything else, we return.

bpf/pca.h
```diff
@@ -0,0 +1,106 @@
...
+       return TC_ACT_UNSPEC;	
+    }   
+    
+    ip = data + sizeof(*eth);
+    if ((void *)ip + sizeof(*ip) > data_end) {
+       return TC_ACT_UNSPEC;
+    }
+
+    if (validate_pca_filter(ip->protocol, (void *)ip + sizeof(*ip), data_end )){
+        return attach_packet_payload(data, data_end, skb, current_time);
```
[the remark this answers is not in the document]
shach33: Sure. Fixed.

bpf/maps_definition.h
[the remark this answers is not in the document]
shach33: done.

bpf/pca.h
```diff
@@ -0,0 +1,106 @@
...
+int ingress_pca_parse (struct __sk_buff *skb) {
+    return export_packet_payload(skb);
+}
+
+SEC("tc_pca_egress")
+int egress_pca_parse (struct __sk_buff *skb) {
+    return export_packet_payload(skb);
+}
+
+#endif /* __PCA_H__ */
```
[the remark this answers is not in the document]
shach33: Added

bpf/pca.h
shach33: Updated.

cmd/netobserv-ebpf-agent.go
```diff
@@ -39,22 +53,30 @@ func main() {
 
 	logrus.WithField("configuration", fmt.Sprintf("%#v", config)).Debugf("configuration loaded")
 
-	flowsAgent, err := agent.FlowsAgent(&config)
-	if err != nil {
-		logrus.WithError(err).Fatal("can't instantiate NetObserv eBPF Agent")
-	}
+	if config.EnablePCA {
+		if config.PCAFilters == "" {
+			logrus.Fatal("[PCA] NetObserv eBPF Agent instantiated without filters to identify packets. Please set environment variable PCA_FILTER.")
```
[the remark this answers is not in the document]
shach33: Should we treat this for capture everything? That might be extremely resource intensive. This was a design decision we made.  This could be updated if the team deems this behavior correct.

pkg/ebpf/tracer.go
shach33: Sure Joel.

cmd/netobserv-ebpf-agent.go
shach33: This was to reuse the same code block for terminating both netobserv and PCA.

bpf/pca.h
shach33: I am not sure what is happening here, but when I experiment, the requests/response generated via dig google.com `+tcp` are also UDP packets.

msherif1234: no that is indeed tcp dns I have used it

cmd/netobserv-ebpf-agent.go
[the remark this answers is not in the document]
msherif1234: wanted to share one of the PM customers asked use case is to do capture all (just learnt this now) cc'd @jpinsonneau

bpf/pca.h
```diff
@@ -0,0 +1,106 @@
...
+        if (packetSize < skb->len){
+            packetSize = skb->len;
+            if (bpf_skb_pull_data(skb, skb->len)){
+                return TC_ACT_UNSPEC;
+            };
+        }
+        // Set flag's upper 32 bits with the size of the paylaod and the bpf_perf_event_output will 
+        // attach the specified amount of bytes from packet to the perf event
+        // https://github.com/xdp-project/xdp-tutorial/tree/9b25f0a039179aca1f66cba5492744d9f09662c1/tracing04-xdp-tcpdump       
+        flags |= (__u64)packetSize << 32;
```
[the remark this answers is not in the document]
shach33: If the size is too big, the call to `bpf_perf_event_output` will fail and return -EFAULT.

pkg/agent/PacketsAgent.go
```diff
@@ -0,0 +1,220 @@
...
+
+	status Status
+}
+
+// ebpfPacketFetcher abstracts the interface of ebpf.FlowFetcher to allow dependency injection in tests
+type ebpfPacketFetcher interface {
+	io.Closer
+	Register(iface ifaces.Interface) error
+
+	LookupAndDeleteMap() map[int][]*byte
```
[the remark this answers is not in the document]
shach33: I am.  This is the one being used to delete map.

pkg/flow/packet_record.go
```diff
@@ -0,0 +1,55 @@
...
+
+// ReadRawPacket reads a PacketRecord from a binary source, in LittleEndian order
+func ReadRawPacket(reader io.Reader) (*PacketRecord, error) {
+	var pr PacketRecord
+	currentTime := time.Now()
+	monotonicTimeNow := monotime.Now()
+	getLen := make([]byte, 4)
+	packetTimestamp := make([]byte, 8)
+	// Read IfIndex and discard it: To be used in other usecases
+	_ = binary.Read(reader, binary.LittleEndian, make([]byte, 4))
```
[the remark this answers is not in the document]
shach33: PCA eventually only transmits a packet stream.  How do you think we should include the interface information?

bpf/pca.h
```diff
@@ -0,0 +1,109 @@
...
+
+static int attach_packet_payload(void *data, void *data_end, struct __sk_buff *skb){
+        payload_meta meta;
+        u64 flags = BPF_F_CURRENT_CPU;
+        // Enable the flag to add packet header
+        // Packet payload follows immediately after the meta struct
+        u32 packetSize = (__u32)(data_end-data);
+        
+        // Record the current time.
+        __u64 current_time = bpf_ktime_get_ns();
```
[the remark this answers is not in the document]
shach33: Fixed.

pkg/flow/perfbuffer.go
```diff
@@ -0,0 +1,67 @@
...
+				plog.Debug("exiting perfbuffer routine")
+				return
+			}
+			if len(c.entries) >= c.maxEntries {
+				evictingEntries := c.entries
+				c.entries = []*PacketRecord{}
+				logrus.WithField("packets", len(evictingEntries)).
+					Debug("evicting packets from userspace accounter after reaching cache max length")
+				c.evict(evictingEntries, out)
+			}
```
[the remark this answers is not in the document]
shach33: It is a debug log. But removed it for now.

bpf/pca.h
shach33: Then, I am able to capture those.

bpf/pca.h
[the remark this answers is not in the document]
shach33: I added SCTP.  ICMP packets dont really have port associated to them. So, it would require an update to the filter specifications, which I can do as part of an extension, but may be not with the current PR.

cmd/netobserv-ebpf-agent.go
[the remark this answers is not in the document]
jpinsonneau: +1 it seems to be a use case now :upside_down_face: 
Would it work if we capture everything expect our own netobserv traffic ? Let's say with unlimited resources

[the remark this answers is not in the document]
shach33: Updated to collect and export all packets if the filter (enviroment variable PCA_FILTER) is not set at all. 
However if the filter is wrongly set, incorrect format/ non-supported protocols etc, then the agent doesnt export anything.

jotak (approved):
Thanks @shach33 , LGTM on my side, I'm leaving it in @msherif1234 / @dushyantbehl hands! :)

bpf/pca.h
```diff
@@ -0,0 +1,121 @@
...
+    else if (ipproto == IPPROTO_SCTP){
+        struct sctphdr *sctp_header = ipheaderend;
+        if ((void *)sctp_header + sizeof(*sctp_header) > data_end) {
+                return false;	
+            }
+            sourcePort = sctp_header->source;
+            destPort = sctp_header->dest;        
+    }
+    else 
+        return false;
```
dushyantbehl, line 76: nit: Please add `{}` here with `else` to keep the code structure same across all C code.

@shach33

bpf/pca.h
```diff
@@ -0,0 +1,121 @@
...
+    if (sourcePort == bpf_htons(pca_port) || destPort == bpf_htons(pca_port)){
```
dushyantbehl, line 77: @shach33  nit: you can possibly do `bpf_htons(pca_port)` once.

dushyantbehl (commented):
Add 2 minor things for @shach33. 
Let me know when they are done and I can add a `lgtm` here. 

Will leave up to @msherif1234 for taking a final view on ebpf and userspace.

shach33:
@msherif1234 This is fixed now.

PR #114, opened 2023-04-25
NETOBSERV-753: Reset ticker if flows flushed when cacheMaxFlows is reached

msherif1234:
ebpf agent optimization to avoid unnecessary flows eviction when `cacheMaxFlows` is reached followed by evictTick timer expiry


jotak (approved):

jotak:
Thanks  @msherif1234 !
/approve

PR #115, opened 2023-04-26
NETOBSERV-1061: Add TCP drop and DNS tracking hooks

msherif1234:
- Add skb_free tracepoint hook to detect when TCP flows are dropped and update flow metrics with tcp socket info.

- To view ebpf tracepoints with `bpftool`
```shell
bpftool perf show
pid 119418  fd 8: prog_id 134  tracepoint  kfree_skb
```
- For reference about tcpdrop https://www.brendangregg.com/blog/2018-05-31/linux-tcpdrop.html

- Steps for manual testing using flowsdump collector tool
```shell
- using ncat
=========
server: ncat -l 8080
client: ncat 192.168.122.69 8080
```

- use iptables rule to drop
============================
```shell
sudo iptables -I INPUT 1 -m tcp --proto tcp --dst 192.168.122.69/32 --dport 8080 -j DROP
switch back
sudo iptables -I INPUT 1 -m tcp --proto tcp --dst 192.168.122.69/32 --dport 8080 -j ACCEPT
```

once the drop iptable rule is installed we can see tcpdrop stats been updated

```sh
 },{
                "cpu": 2,
                "value": {
                    "packets": 1,
                    "bytes": 67,
                    "start_mono_time_ts": 5481943029602,
                    "end_mono_time_ts": 5481943029602,
                    "flags": 16,
                    "errno": 0,
                    "tcp_drops": {
                        "packets": 14,
                        "bytes": 800,
                        "flags": 16,
                        "state": 1,
                        "drop_cause": 8  <<<<< SKB_DROP_REASON_NETFILTER_DROP
                    }
                }	
            }
```
- add `net_dev_queue` trace point hook to implement light weight DNS tracker
- testing using `dig google.coom`
```sh
DNS id 23099
Query
 "dns_record": {
                "id": 23099,
                "flags": 288,
                "req_mono_time_ts": 3559391813403,
                "rsp_mono_time_ts": 0
}

Response
 "dns_record": {
                "id": 23099,
                "flags": 33152,
                "req_mono_time_ts": 0,
                "rsp_mono_time_ts": 3559392215340
}
~.4 msec latency
```
NOTE: 
- here are all possible TCP state values https://elixir.bootlin.com/linux/v6.3/source/include/net/tcp_states.h#L12
- here are all drop causes
https://elixir.bootlin.com/linux/latest/source/include/net/dropreason.h#L88	

Related PRs:
https://github.com/netobserv/network-observability-operator/pull/331
https://github.com/netobserv/flowlogs-pipeline/pull/429
https://github.com/netobserv/network-observability-console-plugin/pull/324


jpinsonneau:
@msherif1234 I tried to deploy this on a clusterbot using `launch 4.13 aws,large` but got the following error:
```
time="2023-04-28T07:46:46Z" level=debug msg="agent IP: 10.0.149.138" component=agent.Flows
time="2023-04-28T07:46:47Z" level=info msg="failed to attach the BPF program to kfree_skb tracepoint: reading file \"/sys/kernel/debug/tracing/events/skb/kfree_skb/id\": open /sys/kernel/debug/tracing/events/skb/kfree_skb/id: no such file or directory" component=ebpf.FlowFetcher
time="2023-04-28T07:46:47Z" level=fatal msg="can't instantiate NetObserv eBPF Agent" error="reading file \"/sys/kernel/debug/tracing/events/skb/kfree_skb/id\": open /sys/kernel/debug/tracing/events/skb/kfree_skb/id: no such file or directory"
```

I thought using `privileged` option in the CRD would solve it but it's not :crying_cat_face: 
```
apiVersion: flows.netobserv.io/v1beta1
kind: FlowCollector
...
spec:
  agent:
    ebpf:
...
      privileged: true
...
```

Am I missing something ?

msherif1234:
> @msherif1234 I tried to deploy this on a clusterbot using `launch 4.13 aws,large` but got the following error:
> 
> ```
> time="2023-04-28T07:46:46Z" level=debug msg="agent IP: 10.0.149.138" component=agent.Flows
> time="2023-04-28T07:46:47Z" level=info msg="failed to attach the BPF program to kfree_skb tracepoint: reading file \"/sys/kernel/debug/tracing/events/skb/kfree_skb/id\": open /sys/kernel/debug/tracing/events/skb/kfree_skb/id: no such file or directory" component=ebpf.FlowFetcher
> time="2023-04-28T07:46:47Z" level=fatal msg="can't instantiate NetObserv eBPF Agent" error="reading file \"/sys/kernel/debug/tracing/events/skb/kfree_skb/id\": open /sys/kernel/debug/tracing/events/skb/kfree_skb/id: no such file or directory"
> ```
> 
> I thought using `privileged` option in the CRD would solve it but it's not 😿
> 
> ```
> apiVersion: flows.netobserv.io/v1beta1
> kind: FlowCollector
> ...
> spec:
>   agent:
>     ebpf:
> ...
>       privileged: true
> ...
> ```
> 
> Am I missing something ?

Yes similar to what I added to e2e manifests we need to add the same mount volume to the crd

eranra:
@msherif1234 I really like this PR/feature. We heard multiple times that customers are looking for this.
 
@msherif1234 In conversations we had in the past with Marc Curry (PM), the thinking was that if we could go one step forward and correlate the drops with a specific "reason" ( the IP tables and maybe higher-level rules set by customers to create those drops), it will provide huge value. I wonder if this needs to be part of the eBPF/Agent work or if we can do this correlation in the user space/FLP etc. @jotak ^^^

jotak:
> @msherif1234 I really like this PR/feature. We heard multiple times that customers are looking for this.
> 
> @msherif1234 In conversations we had in the past with Marc Curry (PM), the thinking was that if we could go one step forward and correlate the drops with a specific "reason" ( the IP tables and maybe higher-level rules set by customers to create those drops), it will provide huge value. I wonder if this needs to be part of the eBPF/Agent work or if we can do this correlation in the user space/FLP etc. @jotak ^^^

I'm also a big fan of it :-)

@msherif1234 have you been able to check if the hook is able to report OVN's dropped traffic from network policies? Even if we're not yet doing the full correlation, this would already be a big step forward.

msherif1234:
> > @msherif1234 I really like this PR/feature. We heard multiple times that customers are looking for this.
> > @msherif1234 In conversations we had in the past with Marc Curry (PM), the thinking was that if we could go one step forward and correlate the drops with a specific "reason" ( the IP tables and maybe higher-level rules set by customers to create those drops), it will provide huge value. I wonder if this needs to be part of the eBPF/Agent work or if we can do this correlation in the user space/FLP etc. @jotak ^^^
> 
> I'm also a big fan of it :-)
> 
> @msherif1234 have you been able to check if the hook is able to report OVN's dropped traffic from network policies? Even if we're not yet doing the full correlation, this would already be a big step forward.

yeah this is what I am working on, I have seen it report drops for iptables rules I inserted for testing now trying with ovs drop BTW there is a bug in ovs to migrate all drops to use skb_free it seems not all drop paths are covered

bpf/flow.h
```diff
@@ -30,8 +32,17 @@ typedef struct flow_metrics_t {
     // 0 otherwise
     // https://chromium.googlesource.com/chromiumos/docs/+/master/constants/errnos.md
     u8 errno;
+    struct tcp_drops_t {
```
jotak, line 35: wouldn't it be more consistent to move `tcp_drops_t` in `tcp_drops.h` ?

bpf/utils.h
```diff
@@ -0,0 +1,278 @@
...
+
+static inline long tcp_drop_lookup_and_update_flow(struct sk_buff *skb, flow_id *id, u8 state, u16 flags,
+                                            enum skb_drop_reason reason) {
+     flow_metrics *aggregate_flow = bpf_map_lookup_elem(&aggregated_flows, id);
+     if (aggregate_flow != NULL) {
+         aggregate_flow->tcp_drops.packets += 1;
+         aggregate_flow->tcp_drops.bytes += skb->len;
+         aggregate_flow->tcp_drops.state = state;
+         aggregate_flow->tcp_drops.flags = flags;
+         aggregate_flow->tcp_drops.drop_cause = reason;
```
jotak, line 268: is bitwise OR possible for these, to accumulate values? I think at least for flags we can do it.
Also I wonder if we could prioritize drop reasons, e.g. if there can be unknown reasons vs known reason we would privilege the known one? does that make sense?

bpf/flow.h
jotak: ah sorry I just realize it's nested

bpf/utils.h
msherif1234: we already filter out unknonwn reasons before we come to this function, regarding the flags yeah I think we can or

pkg/ebpf/tracer.go
```diff
@@ -82,20 +85,30 @@ func NewFlowFetcher(
 		return nil, fmt.Errorf("loading and assigning BPF objects: %w", err)
 	}
 
+	var tcpDropsLink link.Link
+	if tcpDrops {
+		tcpDropsLink, err = link.Tracepoint("skb", "kfree_skb", objects.KfreeSkb, nil)
+		if err != nil {
+			log.Printf("failed to attach the BPF program to kfree_skb tracepoint: %v", err)
```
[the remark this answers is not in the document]
msherif1234: yeah I will use fmt.Errorf and return it

bpf/flow.h
msherif1234: also all structures that needs go eqv need to be in the same .h file

bpf/dns_tracker.h
```diff
@@ -0,0 +1,141 @@
+/*
+    light weight DNS tracker using trace points.
+*/
+
+#ifndef __DNS_TRACKER_H__
+#define __DNS_TRACKER_H__
+#include "utils.h"
+
+#define DNS_PORT        53
+#define DNS_QR_FLAG     0x80
```
tohojo, line 10: This is wrong. The QR flag is the top-most bit in a 16-bit value, so it should be `0x8000`. This value happens to work on a little-endian system because you're not doing ntohs() before the compare, but it'll break on big-endian...

bpf/dns_tracker.h
```diff
@@ -0,0 +1,141 @@
...
+        bpf_probe_read(&dns, sizeof(dns), (struct dns_header *)(skb->head + skb->transport_header + len));
+
+        struct dns_request_key req_key = {
+            .id = dns.id
+        };
+        struct dns_request_val dns_val;
+
+        __builtin_memset(&dns_val, 0, sizeof(dns_val));
+
+        if ((dns.flags & DNS_QR_FLAG) == 0) { /* dns query */
```
tohojo, line 115: See above cf endianness, you should be wrapping the flags value in `bpf_ntohs()`

bpf/dns_tracker.h
```diff
@@ -0,0 +1,141 @@
...
+        len = set_key_with_tcp_info(skb, &id, IPPROTO_TCP, &flags);
+        break;
+    default:
+        return -1;
+    }
+
+    // check for DNS packets
+    if (id.dst_port == DNS_PORT || id.src_port == DNS_PORT) {
+        struct dns_header dns;
+        bpf_probe_read(&dns, sizeof(dns), (struct dns_header *)(skb->head + skb->transport_header + len));
```
tohojo, line 106: This will fail if the DNS packet data is not in the linear part of the skb. Which, depending on the source of the packet may of may not be the case; so realistically you're probably going to miss some requests this way. Also, for TCP in particular, you have no way of checking whether the packet you're looking at is actually the start of the request/response or not, so you may be parsing something that's in the middle of a DNS message as the header and get completely bogus values...

bpf/dns_tracker.h
```diff
@@ -0,0 +1,141 @@
...
+        if ((dns.flags & DNS_QR_FLAG) == 0) { /* dns query */
+            if (bpf_map_lookup_elem(&dns_requests, &req_key) == NULL) {
+                __builtin_memcpy(dns_val.srcip, id.src_ip, IP_MAX_LEN);
+                bpf_map_update_elem(&dns_requests, &req_key, &dns_val, BPF_ANY);
+                find_or_create_dns_flow(&id , &dns, skb->len, INGRESS, flags);
+            }
+        } else { /* dns response */
+            struct dns_request_val *value = bpf_map_lookup_elem(&dns_requests, &req_key);
+            if (value != 0 && (dns_addr_compare(value->srcip, id.dst_ip, IP_MAX_LEN) == 0)) {
+                find_or_create_dns_flow(&id , &dns, skb->len, EGRESS, flags);
```
tohojo, line 124: You're not actually doing anything with the query of the response other than recording the fact that a query with this ID appeared? Couldn't people already basically infer the presence of DNS traffic from just the port numbers? What's the use case for all of this?

bpf/dns_tracker.h
```diff
@@ -0,0 +1,141 @@
...
+            }
+        } // end of dns response
+    } // end of dns port check
+
+    return 0;
+}
+
+SEC("tracepoint/net/net_dev_queue")
```
tohojo, line 132: This means you'll most likely see each packet multiple times; I assume that's fine for the use case?

bpf/dns_tracker.h
```diff
@@ -0,0 +1,141 @@
...
+    flow_id id;
+    u8 protocol = 0;
+    u16 family = 0,flags = 0, len = 0;
+
+    __builtin_memset(&id, 0, sizeof(id));
+
+    // pull in details from the packet headers and the sock struct
+    bpf_probe_read(&family, sizeof(u16), &sk->__sk_common.skc_family);
+
+    id.if_index = skb->skb_iif;
```
tohojo, line 84: I'm not sure if this will be set for a locally-generated packet (or to what value in this case)?

bpf/dns_tracker.h
```diff
@@ -0,0 +1,141 @@
...
+static inline int trace_dns(struct sk_buff *skb) {
+    struct sock *sk = skb->sk;
+    flow_id id;
+    u8 protocol = 0;
+    u16 family = 0,flags = 0, len = 0;
+
+    __builtin_memset(&id, 0, sizeof(id));
+
+    // pull in details from the packet headers and the sock struct
+    bpf_probe_read(&family, sizeof(u16), &sk->__sk_common.skc_family);
```
tohojo, line 82: Not all packets will have a socket associated with them, in which case this will fail. If you just want to extract the packet protocol type, you'll probably be better off looking at the `skb->protocol` field. You'll have to deal with VLAN tagging, though; see the `skb_protocol()` helper in the kernel source...

tohojo (commented):
A few comments on the DNS parsing. The most serious thing is probably the fact that you may miss some data, and that you don't know if you're in the middle of a TCP flow. So while this will probably work fine in the common case, it's likely to result in some weird failure modes in the corner cases. Which will be worse if you're also planning to add parsing of the actual DNS query contents, of course...

bpf/dns_tracker.h
msherif1234: this tool measure the time latency between DNS req till we get response for the same DNS id something very similar to what u get with `dnsbench` tool but it records the latency per flow

bpf/dns_tracker.h
msherif1234: we look for the 1st DNS query then wait for matching response then delete that record, so if we see multiple DNS reqq we will just take the 1st one and ignore the rest

bpf/dns_tracker.h
msherif1234: we don't deal with locally generated packets we handle pkts that enter and exists the container interfaces

bpf/dns_tracker.h
tohojo: Right, there's no guarantee you won't get collisions, though? ATAICT you're not using the src/dst IP and port as part of the DNS key, so you just have the 16-bit DNS ID as a globally unique identifier? 16 bits is not a lot, so you'll definitely get collisions this way...

Also, is there a mechanism for cleaning out the map of entries that never get a response somewhere?

bpf/dns_tracker.h
tohojo: How are you ensuring that?

bpf/dns_tracker.h
tohojo: Right, OK.

msherif1234:
> A few comments on the DNS parsing. The most serious thing is probably the fact that you may miss some data, and that you don't know if you're in the middle of a TCP flow. So while this will probably work fine in the common case, it's likely to result in some weird failure modes in the corner cases. Which will be worse if you're also planning to add parsing of the actual DNS query contents, of course...

What are ur suggestions to make this better with TCP do I need to check tcpflags ?

tohojo:
> What are ur suggestions to make this better with TCP do I need to check tcpflags ?

Well, there's fundamentally no way to know, from looking at a single TCP packet, where in the flow that packet is. You'd have to implement a complete TCP flow state tracker, which I would definitely not recommend for this!

You could probably do something to alleviate the chance of false matches, though. Something like:
- Make sure you include the IP 5-tuple into the lookup when looking up the DNS request ID (you probably want to do this anyway, see my comment about collisions above)
- When dealing with TCP traffic, make sure you only ever accept a new DNS request entry the first time you see one *on that 5-tuple*.
- When dealing with a DNS *reply*, never accept it unless there's a matching request.

This will not be completely "correct", still, but I think it will probably be good enough to minimise the number of false positives you'll see in practice...

bpf/dns_tracker.h
msherif1234: its LRU map so I assume the stale entry will get magically cleared out by kernel no ?

msherif1234:
cpu and memory usage after turning on DNS tracking
image

bpf/dns_tracker.h
msherif1234: fixed

bpf/dns_tracker.h
msherif1234: fixed

bpf/dns_tracker.h
msherif1234: fixed

bpf/utils.h
jotak: but like we do `aggregate_flow->tcp_drops.bytes += skb->len`, which means basically we add the input data into the aggregate object, why don't we do the same for flags: `aggregate_flow->tcp_drops.flags |= flags;` (instead of just `=`) ?

msherif1234: will fix it

jotak (dismissed):
Thanks @msherif1234 
this lgtm - as said I'm mostly reviewing the user space part as I'm less legit on the C code

bpf/dns_tracker.h
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    // check for DNS packets
+    if (id.dst_port == DNS_PORT || id.src_port == DNS_PORT) {
+        struct dns_header dns;
+        bpf_probe_read(&dns, sizeof(dns), (struct dns_header *)(skb->head + skb->transport_header + len));
+        if ((bpf_ntohs(dns.flags) & DNS_QR_FLAG) == 0) { /* dns query */
+            id.direction = INGRESS;
+        } else { /* dns response */
+            id.direction = EGRESS;
```
tohojo, line 86: Ingress and egress are a bit arbitrary for something like this, of course, but I found it surprising that the request is set to ingress and the reply to egress? Aren't DNS queries mostly going out of the machine/pod?

bpf/flow.h
```diff
@@ -30,8 +32,24 @@ typedef struct flow_metrics_t {
...
+    struct dns_record_t {
+        u16 id;
+        u16 flags;
+        u64 req_mono_time_ts;
+        u64 rsp_mono_time_ts;
+    } __attribute__((packed)) dns_record;
 } __attribute__((packed)) flow_metrics;
 
+// Force emitting struct tcp_drops into the ELF.
+const struct tcp_drops_t *unused0 __attribute__((unused));
```
tohojo, line 51: Why are these needed?

bpf/utils.h
```diff
@@ -0,0 +1,299 @@
...
+
+static inline long tcp_drop_lookup_and_update_flow(struct sk_buff *skb, flow_id *id, u8 state, u16 flags,
+                                            enum skb_drop_reason reason) {
+     flow_metrics *aggregate_flow = bpf_map_lookup_elem(&aggregated_flows, id);
+     if (aggregate_flow != NULL) {
+         aggregate_flow->tcp_drops.packets += 1;
+         aggregate_flow->tcp_drops.bytes += skb->len;
+         aggregate_flow->tcp_drops.state = state;
+         aggregate_flow->tcp_drops.flags |= flags;
+         aggregate_flow->tcp_drops.drop_cause = reason;
```
tohojo, line 289: Having these variables in an "aggregate flow" entry seems a bit weird. So the `state` and `drop_reason` fields will be set to the last seen socket state and drop cause for the flow, and the flags will be a bitwise OR of all packet flags ever seen on the connection? Just overriding values with the last value seen seems like a good way to lose data; and the bitwise combination of the flags just doesn't make any semantic sense. What are there fields actually meant to be used for?

bpf/flow.h
```diff
@@ -10,6 +10,8 @@ typedef __u16 u16;
 typedef __u32 u32;
 typedef __u64 u64;
```
tohojo, line 11: The convention for most BPF code is to use the underscored values instead of type-def'ing like this. Is there any particular reason why you went with these typedefs?

bpf/dns_tracker.h
msherif1234: this took us to dns pod where queries are input and respones were output but as u said this arbitrary since skb_buff isn't direction aware 
image
do u prefer if I switched the direction ?

bpf/flow.h
msherif1234: this is cilium magic to generate go structures from c structures for the userspace

bpf/flow.h
msherif1234: no real reason that was the style used in that file and I just kept it as is

bpf/flow.h
tohojo: Ah, right, gotcha!

bpf/utils.h
tohojo: Well, depends. What do you do when you get multiple drops on the same flow (that is quite normal in a TCP session)? I can think of two options:
- Always create a new flow entry for each drop.
- Keep the flow entry, but have multiple drop events associated with each flow

From a protocol perspective, the latter is more correct (logically, the multiple drops would be associated with the same flow), and you could have a drop counter for a flow that would increment, etc. You'll need to rework the data structure so you can accommodate that, though...

bpf/dns_tracker.h
msherif1234: I will switched around then

bpf/utils.h
tohojo: Well that's the thing that's by no means clear to me (whether the last drop reason is the important one). But it that's the use case, fair enough; definitely get rid of the `|=` on the flags field, though (or just lose the field entirely)...

msherif1234: As user I see tcp connection is failing and wanted to know why so I will remove the or we might have interest to know what was the tcp flags when drop occurs

jotak: Hey,
I did also discuss this with @msherif1234 - I agree that it is quite arbitrary to show only the last reason, but there's also a tradeoff to do between debugability and complexity (and performance impact - all events will be stored), so it seems reasonable to me to only pick 1 reason; that said, I cannot tell if the last would be statistically more significant.. maybe the first would be more? Hard to say, perhaps yes. Any thought about that?
Keeping all flags was easier as it doesn't involve to re-shape the data model.
Perhaps to avoid confusion, the `drop_cause` should be renamed `last_drop_cause` ?

bpf/dns_tracker.h
jotak: @msherif1234 side question: how does that work with external DNS? Is this feature limited to in-cluster dns ?

bpf/utils.h
msherif1234: I think keeping history of all drops for older transient drops will just add confusion without much value imo I am fine with rename if that will make it more clear

tohojo: I don't think we can predict, in general, which drop will be the most significant to debug any particular issue. It really depends on what kind of connection issue you're trying to debug; I *suspect* having the full drop history will turn out to be useful in the future, but I really don't have anything other than a feeling to back that up with, so having just the last reason is probably a reasonable start... :) Renaming the field so anyone looking at it can tell that it's in fact the last reason is a good idea, though!

To expand a bit on why I think OR'ing the flags field is a bad idea: the issue is basically that the flags are not exclusive. E.g., the first packet in a connection will have the SYN flag set, the second one will have SYN+ACK, and subsequent packets will have just the ACK packet set. So say you look at the data and you see the drop counter has a value of 2, and the flags field is set to SYN+ACK. Does that mean that the last packet dropped was a SYN+ACK packet, or was there an earlier SYN packet that got dropped, retransmitted, and later a single ACK got dropped? You have no way to tell because of the OR'ing. So better to keep just the last flags so you can unambiguously refer that to the drop reason; and if you need the full history of drops for a connection, add proper history tracking with separate events per drop.

jotak: Well, we have a trick here:
```c
// Custom flags exported
#define SYN_ACK_FLAG 0x100
#define FIN_ACK_FLAG 0x200
#define RST_ACK_FLAG 0x400
```
:-)

But anyway, that's fine with me to keep latest flag only! We don't have any strong requirement about that

bpf/utils.h
```diff
@@ -0,0 +1,298 @@
...
+    // ICMPv4/ICMPv6 code value
+    u8 icmp_code;
+    // TCP flags
+    u16 flags;
+};
+
+const u8 ip4in6[] = {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0xff, 0xff};
+
+// Constant definitions, to be overridden by the invoker
+volatile const u32 sampling = 0;
```
dushyantbehl, line 37: Does it make sense to move the constants which act more like a configuration variable enable/disable switch to be present in either a separate file `config.h` or on top of the main file for visibility?

bpf/utils.h
```diff
@@ -0,0 +1,298 @@
...
+     bpf_probe_read(&eth, sizeof(eth), (struct ethhdr *)(skb->head + skb->mac_header));
+     id->eth_protocol = bpf_ntohs(eth.h_proto);
+     __builtin_memcpy(id->dst_mac, eth.h_dest, ETH_ALEN);
+     __builtin_memcpy(id->src_mac, eth.h_source, ETH_ALEN);
+    if (id->eth_protocol == ETH_P_IP) {
+        *family = AF_INET;
+    } else if (id->eth_protocol == ETH_P_IPV6) {
+        *family = AF_INET6;
+    }
+ }
```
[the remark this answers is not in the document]
msherif1234: sure we can do so

bpf/utils.h
msherif1234: I think I can move those two the flows.c file as it was to make it visible ?

dushyantbehl: yes for visibility to easily track them I meant.

msherif1234: I will move them today

dushyantbehl (approved):
/lgtm

bpf/dns_tracker.h
jotak: Sorry for digging up this old comment; @msherif1234 / @tohojo do you see some sanity check that we could do to ensure this is a valid dns header?

Issue #102, opened 2023-03-09
NETOBSERV-1112: Support adding of connection setup timestamp


praveingk:
Currently the flow records from ebpf agent provide two timestamps : 1) connection start, and 2) connection end.
Additionally, we could have connection setup timestamp, which is either (SYN+ACK), or ACK+Seq1 timestamp depending on the direction of the flow.

The logic could : 

**Ingress:** 
1) Store timestamp (t2) when ACK+SEQ1 packet is received. Incoming SYN packet timestamp (t1) is already stored.
2) Store timestamp (t2) when SYN+ACK packet is received. Outgoing SYN packet timestamp (t1) would already be stored. Mostly this case is already captured in eBPF data path, since we store the timestamp when an entry is created in the map.  In this case SYN+ACK packet will create a new entry in the map.

Using existing marker stored (for SYN+ACK, SYN) in TCP flags, we can calculate t2-t1 in user space to get connection setup time.

This connection setup RTT can be captured per flow, and exported. The consumer can then estimate the Path RTT based on the moving average or some other calculation.

<img width="394" alt="image" src="https://user-images.githubusercontent.com/9940568/223928774-6f81048c-7ebc-4160-b22a-5a8b5d31f5e4.png">

CC : @eranra @jotak @msherif1234

jotak:
sounds good to me!
@praveingk is that something you want to work on? (I see you took the assignment).

praveingk:
@jotak : Yes, I was planning to work on it. If someone else wants to work on it, I am ok wth that. 👍🏽

dushyantbehl:
@praveingk @jotak @msherif1234 @eranra 

While working on this issue I noticed a problem with the above design. The hypothesis above assumes that the packets start with sequence number 0 for SYN and hence we can record the timestamp of SYN+ACK and ACK+SEQ1 in the agent and perform RTT calculation in the userspace.
The actual TCP behaviour is slightly different because TCP standard starts packets with a random sequence number from both side like below,

<img width="394" alt="packet flow" src="https://user-images.githubusercontent.com/1579780/234001141-72c6c5d2-a12e-4610-b6b1-d72fcf557ed7.png">

[...]

PR #117, opened 2023-04-28
NETOBSERV-1112: Add TCP RTT calculation to ebpf-agent and userspace.

dushyantbehl:
Fixes https://github.com/netobserv/netobserv-ebpf-agent/issues/102

Feature is disabled by default and can be enable with an environment variable flag.

Here are the CPU and Memory Numbers before and after the PR

Screenshot 2023-06-30 at 11 10 51 AM

Screenshot 2023-06-30 at 11 11 07 AM


dushyantbehl:
Tested on my end to be working end to end for calculating RTT for TCP handshake.

eranra:
@dushyantbehl, thanks for working on this PR ( and also, thanks, @praveingk ). @dushyantbehl Can you add some text that explains the logic of the calculation? I suggest using `.md` file and keep under docs. This will help both to review + for the future to explain what is happening.

eranra:
@jotak @jpinsonneau 
the agent reports RTT just at the beginning of flows --- so all future/other flows in the conversation (with the same L4 hash) do not get the info. I wonder what we want to show in the flows table? maybe we need to 'clone' the RTT info in all flows from the same conversation? (but this is not really measured information)

eranra:
@dushyantbehl I ran some high-level test, and the results are ok, but not exactly what I expected::

(1) I ran in shell#1 `while true; do wget http://hgd-speedtest-1.tele2.net/1MB.zip; sleep 10; done`
This downloads a file from a known test server 

(2) I ran in shell#2 `sudo FLOWS_TARGET_HOST=127.0.0.1 FLOWS_TARGET_PORT=9999 ./bin/netobserv-ebpf-agent`
starts the agent 

(3) I ran in shell#3 `./bin/flowlogs-dump-collector -listen_port=9999 2>&1 >/dev/null | grep 90.130.74.149`
dump to screen

The output I get is::

```
$ ./bin/flowlogs-dump-collector -listen_port=9999 2>&1 >/dev/null | grep 90.130.74.149
ipv4: 11:15:34.760990 eth0 IP 172.26.89.33:35262 > 90.130.74.149:80: protocol:tcp type: 0 code: 0 dir:1 bytes:20945 packets:315 flags:530 ends: 11:15:35.550359 rtt(us): 0
ipv4: 11:15:34.853181 eth0 IP 90.130.74.149:80 > 172.26.89.33:35262: protocol:tcp type: 0 code: 0 dir:0 bytes:1069618 packets:315 flags:272 ends: 11:15:35.549528 rtt(us): 92190
ipv4: 11:15:35.639099 eth0 IP 90.130.74.149:80 > 172.26.89.33:35262: protocol:tcp type: 0 code: 0 dir:0 bytes:66 packets:1 flags:512 ends: 11:15:35.639099 rtt(us): 0
ipv4: 11:15:35.639127 eth0 IP 172.26.89.33:35262 > 90.130.74.149:80: protocol:tcp type: 0 code: 0 dir:1 bytes:66 packets:1 flags:16 ends: 11:15:35.639127 rtt(us): 0
ipv4: 11:15:45.560058 eth0 IP 172.26.89.33:38364 > 90.130.74.149:80: protocol:tcp type: 0 code: 0 dir:1 bytes:74 packets:1 flags:2 ends: 11:15:45.560058 rtt(us): 0
ipv4: 11:15:45.654908 eth0 IP 90.130.74.149:80 > 172.26.89.33:38364: protocol:tcp type: 0 code: 0 dir:0 bytes:1069354 packets:311 flags:784 ends: 11:15:46.489739 rtt(us): 94850
ipv4: 11:15:45.654979 eth0 IP 172.26.89.33:38364 > 90.130.74.149:80: protocol:tcp type: 0 code: 0 dir:1 bytes:20739 packets:312 flags:528 ends: 11:15:46.489763 rtt(us): 0
ipv4: 11:15:56.404436 eth0 IP 172.26.89.33:47890 > 90.130.74.149:80: protocol:tcp type: 0 code: 0 dir:1 bytes:21077 packets:317 flags:530 ends: 11:15:57.340466 rtt(us): 0
ipv4: 11:15:56.499668 eth0 IP 90.130.74.149:80 > 172.26.89.33:47890: protocol:tcp type: 0 code: 0 dir:0 bytes:1069684 packets:316 flags:784 ends: 11:15:57.340431 rtt(us): 95232
ipv4: 11:16:07.255185 eth0 IP 172.26.89.33:55564 > 90.130.74.149:80: protocol:tcp type: 0 code: 0 dir:1 bytes:20417 packets:307 flags:530 ends: 11:16:08.124477 rtt(us): 0
ipv4: 11:16:07.344960 eth0 IP 90.130.74.149:80 > 172.26.89.33:55564: protocol:tcp type: 0 code: 0 dir:0 bytes:1068958 packets:305 flags:784 ends: 11:16:08.124456 rtt(us): 89774
ipv4: 11:16:18.044159 eth0 IP 172.26.89.33:47140 > 90.130.74.149:80: protocol:tcp type: 0 code: 0 dir:1 bytes:18797 packets:271 flags:530 ends: 11:16:18.967787 rtt(us): 0
ipv4: 11:16:18.133470 eth0 IP 90.130.74.149:80 > 172.26.89.33:47140: protocol:tcp type: 0 code: 0 dir:0 bytes:1067836 packets:288 flags:784 ends: 11:16:18.967766 rtt(us): 89310
ipv4: 11:16:28.982189 eth0 IP 90.130.74.149:80 > 172.26.89.33:53638: protocol:tcp type: 0 code: 0 dir:0 bytes:1069156 packets:308 flags:784 ends: 11:16:29.758358 rtt(us): 87864
ipv4: 11:16:28.894324 eth0 IP 172.26.89.33:53638 > 90.130.74.149:80: protocol:tcp type: 0 code: 0 dir:1 bytes:20615 packets:310 flags:530 ends: 11:16:29.758379 rtt(us): 0
```



When I run `ping hgd-speedtest-1.tele2.net`, I get 

```
PING hgd-speedtest-1.tele2.net (90.130.74.149) 56(84) bytes of data.
64 bytes from hgd-speedtest-1.tele2.net (90.130.74.149): icmp_seq=1 ttl=49 time=88.0 ms
64 bytes from hgd-speedtest-1.tele2.net (90.130.74.149): icmp_seq=2 ttl=49 time=89.6 ms
64 bytes from hgd-speedtest-1.tele2.net (90.130.74.149): icmp_seq=3 ttl=49 time=89.0 ms
64 bytes from hgd-speedtest-1.tele2.net (90.130.74.149): icmp_seq=4 ttl=49 time=88.7 ms
64 bytes from hgd-speedtest-1.tele2.net (90.130.74.149): icmp_seq=5 ttl=49 time=89.7 ms
64 bytes from hgd-speedtest-1.tele2.net (90.130.74.149): icmp_seq=6 ttl=49 time=88.8 ms
```

Now, the rtt 88ms ( from ping) is similar to the reported RTT we get from the eBPF agent :-) This looks good. The thing that is a bit challenging is the direction of the reports and the zeros. I see that we get RTT just in the return path (which makes sense), but it is a bit strange. We need to fix that in the UI or maybe in the user-space agent somehow. Can you add to the dump also the flags so we will know in which flows we get "syn-acks" and ion which reports we do not?

eranra (approved):
It works :-)

jpinsonneau:
> @jotak @jpinsonneau the agent reports RTT just at the beginning of flows --- so all future/other flows in the conversation (with the same L4 hash) do not get the info. I wonder what we want to show in the flows table? maybe we need to 'clone' the RTT info in all flows from the same conversation? (but this is not really measured information)

From the conversation ID we can link back to the first flow and show any missing info. I would be more in favor of that than duplicating data; even if it would require an extra query in some cases.

FYI there is [a doc for conversation tracking followup](https://docs.google.com/document/d/1oZ-UD6kceKpsp5ujeQXK_i8b13SyhT7ka6G4MVguxvM). Feel free to add any thoughts :smile_cat:

jotak:
Thanks @dushyantbehl !

> @jotak @jpinsonneau the agent reports RTT just at the beginning of flows --- so all future/other flows in the conversation (with the same L4 hash) do not get the info. I wonder what we want to show in the flows table? maybe we need to 'clone' the RTT info in all flows from the same conversation? (but this is not really measured information)

Help me understand.. the RTT is the time elapsed between the SYN emission and the ACK reception - so when when you say future flows in the conversation do not get the info, do you mean there could be more SYN/ACK for the same n-tuples? Or do you see other cases of potential loss of information?


Also .. I have a concern, is it safe with sampling? ( @dushyantbehl @eranra @praveingk )
E.g. if the ingress/ACK of the handshake wasn't seen due to sampling, but subsequent ACK are received, is it going to result in screwed up measurements?

bpf/flows.c
```diff
@@ -234,22 +248,98 @@ static inline int fill_ethhdr(struct ethhdr *eth, void *data_end, flow_id *id, u
...
+    __builtin_memset(&seq_id, 0, sizeof(flow_seq_id));
+
+    switch (pkt->id->transport_protocol)
+    {
+    case IPPROTO_TCP: {
+            struct tcphdr *tcp = (struct tcphdr *) pkt->l4_hdr;
+            if ( !tcp || ((void *)tcp + sizeof(*tcp) > data_end) ) {
+                break;
+            }
+            if ((direction == EGRESS) && IS_SYN_PACKET(pkt)) {
```
jotak, line 278: Related to my comment here https://github.com/netobserv/netobserv-ebpf-agent/pull/117#issuecomment-1532669464
I wonder if it would be better to write here:

```c
if (pkt->flags & SYN_FLAG) {
```
and below:

```c
if (pkt->flags & SYN_ACK_FLAG) {
```
(even dropping the direction check)

Since we use a dedicated flag for SYN+ACK ?

dushyantbehl:
> > @jotak @jpinsonneau the agent reports RTT just at the beginning of flows --- so all future/other flows in the conversation (with the same L4 hash) do not get the info. I wonder what we want to show in the flows table? maybe we need to 'clone' the RTT info in all flows from the same conversation? (but this is not really measured information)
> 
> From the conversation ID we can link back to the first flow and show any missing info. I would be more in favor of that than duplicating data; even if it would require an extra query in some cases.
> 
> FYI there is [a doc for conversation tracking followup](https://docs.google.com/document/d/1oZ-UD6kceKpsp5ujeQXK_i8b13SyhT7ka6G4MVguxvM). Feel free to add any thoughts 😸

@jpinsonneau I am not able to access the doc could you please share access with my redhat account, username is `dbehl`.

bpf/flows.c
dushyantbehl: @jotak referencing the figure from my comment here - https://github.com/netobserv/netobserv-ebpf-agent/issues/102#issuecomment-1520129627

We check against these combinations - 
For Client: SYN (Egress) and SYN+ACK (Ingress)
For Server: SYN+ACK (Egress) and ACK(Ingress)

Hence the checks are like this,
```
#define IS_SYN_PACKET(pkt)    ((pkt->flags & SYN_FLAG) || (pkt->flags & SYN_ACK_FLAG))
#define IS_ACK_PACKET(pkt)    ((pkt->flags & ACK_FLAG) || (pkt->flags & SYN_ACK_FLAG))
```

On the second point, I could potentially drop the direction check but that would mean that a node which sends a SYN packet will also check every ACK that it sends to see if it is present in the `flow_sequences` identifier hashmap (would be an extra check for ACKs originating from that node)

dushyantbehl:
> Thanks @dushyantbehl !
> 
> > @jotak @jpinsonneau the agent reports RTT just at the beginning of flows --- so all future/other flows in the conversation (with the same L4 hash) do not get the info. I wonder what we want to show in the flows table? maybe we need to 'clone' the RTT info in all flows from the same conversation? (but this is not really measured information)
> 
> Help me understand.. the RTT is the time elapsed between the SYN emission and the ACK reception - so when when you say future flows in the conversation do not get the info, do you mean there could be more SYN/ACK for the same n-tuples? Or do you see other cases of potential loss of information?

Correct, RTT here is for initial handshake right now, the differnece between the time when SYN packet gets sent and when a corresponding ACK packet is received.
I think I should fix my comment because what I meant was that right now this patch will calculate RTT for handshake packets and not for all packets in the flow so we won't see the RTT measurement for many flow records being evicted from ebpf-agent which correspond to packets other than initial handshake. 

Ideally we should do calculations for all packets because latency can vary in the network at any time which is something we can add on top of this patch. 

> Also .. I have a concern, is it safe with sampling? ( @dushyantbehl @eranra @praveingk ) E.g. if the ingress/ACK of the handshake wasn't seen due to sampling, but subsequent ACK are received, is it going to result in screwed up measurements?

Its safe to the level that this won't cause any wrong RTT values to be shown but if a SYN or its corresponding ACK packet gets sampled we won't see RTT measurement for that flow handshake because any ACK that will come later will have a different sequence id and not the id of first one.

The only problem I foresee is if first ACKs are re-transmitted due to some reason the RTT calculation might not reflect real world round trip latency but in that case I'm not sure how best to mitigate this issue other than perform RTT measurement for all packets in the flow and then take a running average or something on top.

dushyantbehl:
> Now, the rtt 88ms ( from ping) is similar to the reported RTT we get from the eBPF agent :-) This looks good. The thing that is a bit challenging is the direction of the reports and the zeros. I see that we get RTT just in the return path (which makes sense), but it is a bit strange. We need to fix that in the UI or maybe in the user-space agent somehow. Can you add to the dump also the flags so we will know in which flows we get "syn-acks" and ion which reports we do not?

Sure thing, I'm assuming by the dump you mean [flowlogs-dump-collector.go](https://github.com/netobserv/netobserv-ebpf-agent/blob/main/examples/flowlogs-dump/server/flowlogs-dump-collector.go)  it already has the flags just that they are not printed as SYN/ACK/FIN etc but just dumped in binary form. I can change that if that's what you meant.

dushyantbehl:
> @jotak @jpinsonneau the agent reports RTT just at the beginning of flows --- so all future/other flows in the conversation (with the same L4 hash) do not get the info. I wonder what we want to show in the flows table? maybe we need to 'clone' the RTT info in all flows from the same conversation? (but this is not really measured information)

Added @eranra please check

jotak:
FYI: @msherif1234 found issues about the per-cpu map structure, that apparently would have impacts on the RTT calculation (especially if we want RTT for the whole connection lifetime and not just for the handshake) - he has a PR (https://github.com/netobserv/netobserv-ebpf-agent/pull/118) , we need to check the performance impact before taking any decision.

dushyantbehl:
> FYI: @msherif1234 found issues about the per-cpu map structure, that apparently would have impacts on the RTT calculation (especially if we want RTT for the whole connection lifetime and not just for the handshake) - he has a PR (#118) , we need to check the performance impact before taking any decision.

@jotak  Actually the RTT that I am calculating does not get affected by any map decision that we make for the flow logs which we need to pass to user space. This design just calculate RTT when it captures a packet and does not have any map in its path unless you are talking about the map where I store RTT which is separate from the per cpu structure used in ebpf agent currently.
If this is unclear let me setup a meeting with the design of the RTT calculation among the no-dev-team and we can clear this out.

jotak:
we'll hold that until we cut 1.3 branch, then it can be tested / merged (targeting 1.4)
Thanks @dushyantbehl !
/hold

dushyantbehl:
@msherif1234 @jotak @eranra can we move forward with this before #115 it is a minor change and has already been approved before.

dushyantbehl:
@jotak @msherif1234 @eranra 
Added an environment variable which acts as a switch to enable RTT calculation in the ebpf agent. 
Updated docs and configuration file to reflect the same. 
This feature is now `disabled by default`.

/ok-to-test

jotak:
Thanks for adding the enable switch.
Looks good to me, but deferring to @msherif1234 for approval

dushyantbehl:
/hold 

Rebased on top of latest changes. Will unhold tomorrow after doing a dev testing.

dushyantbehl:
@msherif1234 changes have been rebased. Open to review.

dushyantbehl:
@msherif1234 Dev testing is done. The behavior looked as expected. Open to review and merging when approved. 

/unhold /ok-to-test

msherif1234:
> @msherif1234 Dev testing is done. The behavior looked as expected. Open to review and merging when approved.

Thanks!! I will look today

bpf/flows.c
```diff
@@ -102,4 +25,32 @@ int egress_flow_parse(struct __sk_buff *skb) {
...
+
+SEC("tracepoint/net/net_dev_queue")
+int trace_net_packets(struct trace_event_raw_net_dev_template *args) {
+    struct sk_buff skb;
+
+    __builtin_memset(&skb, 0, sizeof(skb));
+    bpf_probe_read(&skb, sizeof(struct sk_buff), args->skbaddr);
+    return trace_dns(&skb);
+}
+
```
msherif1234, line 54: why did move all sec here instead of keeping it in its own file as it was ? because this file will be empty ? maybe that is ok I think its cleaner to keep the implementation with the ist section and `.c` file is just container for all features

bpf/maps_definition.h
```diff
@@ -18,4 +18,17 @@ struct {
...
+// then the hashmap entry will need to be evicted
+// Key is flow_seq_id which is standard 4 tuple and a sequence id
+//     sequence id is specific to the type of transport protocol
+// Value is u64 which represents the occurrence timestamp of the packet.
+struct {
+    __uint(type, BPF_MAP_TYPE_LRU_HASH);
+    __uint(max_entries, 1 << 20);   // Will take around 64MB of space.
+    __type(key, flow_seq_id);
+    __type(value, u64);
+} flow_sequences SEC(".maps");
```
msherif1234, line 32: pls collect cpu and memory profile when the feature is enabled and add it here for reference u can do that by bring up a cluster and check memory and cpu metrics before and after

msherif1234: BTW the way LRU evicts is when the table reach its capacity
```
The BPF_MAP_TYPE_LRU_HASH and BPF_MAP_TYPE_LRU_PERCPU_HASH variants add LRU semantics to their respective hash tables. An LRU hash will automatically evict the least recently used entries when the hash table reaches capacity. An LRU hash maintains an internal LRU list that is used to select elements for eviction. This internal LRU list is shared across CPUs but it is possible to request a per CPU LRU list with the BPF_F_NO_COMMON_LRU flag when calling bpf_map_create.
```

bpf/rtt_tracker.h
```diff
@@ -0,0 +1,69 @@
...
+                    pkt->rtt = pkt->current_ts - *prev_ts;
+                    // Delete the flow from flow sequence map so if it
+                    // restarts we have a new RTT calculation.
+                    long ret = bpf_map_delete_elem(&flow_sequences, &seq_id);
+                    if (trace_messages && ret != 0) {
+                        bpf_printk("error evicting flow sequence: %d", ret);
+                    }
+                }
+                break;
+            }
```
msherif1234, line 62: since this is only for tcp I will check
```C
if (protocol != tcp) {
  return;
}  

switch (dir) {
  case ingress:....
  case egress: ....
}
```

bpf/rtt_tracker.h
```diff
@@ -0,0 +1,69 @@
...
+                if (trace_messages && ret != 0) {
+                    bpf_printk("err saving flow sequence record %d", ret);
+                }
+                break;
+            }
+            if ((direction == INGRESS) && IS_ACK_PACKET(pkt)) {
+                // Stored sequence should be ack_seq - 1
+                u32 seq = bpf_ntohl(tcp->ack_seq) - 1;
+                // check reversed flow
+                fill_flow_seq_id(&seq_id, pkt, seq, 1);
```
msherif1234, line 49: what happens when there are retransimits ?

bpf/flow_monitor.h
```diff
@@ -0,0 +1,128 @@
...
+            }
+            record->id = id;
+            record->metrics = new_flow;
+            bpf_ringbuf_submit(record, 0);
+        }
+    }
+    return TC_ACT_OK;
+}
+
+#endif /* __FLOW_MONITOR_H__ */
```
msherif1234, line 128: nit: pls add new line here

bpf/rtt_tracker.h
```diff
@@ -0,0 +1,69 @@
...
+                }
+                break;
+            }
+        } break;
+    default:
+        break;
+    }
+}
+
+#endif /* __RTT_TRACKER_H__ */
```
msherif1234, line 69: nit: pls add newline here as well

bpf/types.h
```diff
@@ -79,6 +125,16 @@ typedef struct flow_id_t {
 // Force emitting struct flow_id into the ELF.
 const struct flow_id_t *unused2 __attribute__((unused));
 
+// Standard 4 tuple and a sequence identifier.
+// No need to emit this struct. It's used only in kernel space
+typedef struct flow_seq_id_t {
+    u16 src_port;
+    u16 dst_port;
+    u8 src_ip[16];
```
msherif1234, line 133: pls use `IP_MAX_LEN` instead of hardcoded `16`

bpf/types.h
```diff
@@ -93,4 +149,13 @@ const struct flow_record_t *unused3 __attribute__((unused));
...
+// Internal structure: Packet info structure parsed around functions.
+typedef struct pkt_info_t {
+    flow_id *id;
+    u64 current_ts; // ts recorded when pkt came.
+    u16 flags;      // TCP specific
+    void *l4_hdr;   // Stores the actual l4 header
+    u64 rtt;        // rtt calculated from the flow if possible. else zero
+} pkt_info;
+
+#endif /* __TYPES_H__ */
```
msherif1234, line 161: nit new line here

bpf/utils.h
```diff
@@ -137,68 +87,59 @@ static inline void fill_l4info(void *l4_hdr_start, void *data_end, u8 protocol,
...
-    fill_l4info(l4_hdr_start, data_end, ip->nexthdr, &l4_info);
-    id->src_port = l4_info.src_port;
-    id->dst_port = l4_info.dst_port;
-    id->icmp_type = l4_info.icmp_type;
-    id->icmp_code = l4_info.icmp_code;
-    *flags = l4_info.flags;
+    flow_id *id = pkt->id;
+    /* Save the IP Address to id directly. copy once. */
+    __builtin_memcpy(id->src_ip, ip->saddr.in6_u.u6_addr8, 16);
+    __builtin_memcpy(id->dst_ip, ip->daddr.in6_u.u6_addr8, 16);
```
msherif1234, line 120: pls use `IP_MAX_LEN`

pkg/agent/agent.go
```diff
@@ -122,7 +122,7 @@ func FlowsAgent(cfg *Config) (*Flows, error) {
 		debug = true
 	}
 
-	fetcher, err := ebpf.NewFlowFetcher(debug, cfg.Sampling, cfg.CacheMaxFlows, ingress, egress, cfg.EnableTCPDrops, cfg.EnableDNSTracking)
+	fetcher, err := ebpf.NewFlowFetcher(debug, cfg.Sampling, cfg.CacheMaxFlows, ingress, egress, cfg.EnableTCPDrops, cfg.EnableDNSTracking, cfg.EnableRTT)
```
msherif1234, line 125: I am sorry I kind of started this can u pls create a feature structure and pass it as an arg if not I can follow up after ur pr

pkg/agent/config.go
```diff
@@ -136,6 +136,8 @@ type Config struct {
 	KafkaSASLClientSecretPath string `env:"KAFKA_SASL_CLIENT_SECRET_PATH"`
 	// ProfilePort sets the listening port for Go's Pprof tool. If it is not set, profile is disabled
 	ProfilePort int `env:"PROFILE_PORT"`
+	// Enable RTT calculations for the flows, default is 0 (disabled), set to 1 to enable.
```
msherif1234, line 139: `default is false set to true to enable`

bpf/maps_definition.h
dushyantbehl: Did you mean something like a QE run?

bpf/flows.c
dushyantbehl: yes that's what my thought was, after following your refactoring style I felt that this file was getting empty so I thought we can put all the various hookpoints in the .c file for visibility on which features the agent supports (optional or not doesn't matter) and keep the feature specefic implementation in their respective files.

If this is not okay or a clear point I am happy to move things back like before.

bpf/flow_monitor.h
dushyantbehl: Done. Understood the style, will follow it.

bpf/rtt_tracker.h
dushyantbehl: In case of retransmissions we have two scenarios, if SYN is retransmitted the last one will be taken into account which is the correct behavior.
If ACK is retransmitted the receiver will only see the last ACK which it received and for the receiver the calculated RTT will be bigger than the line value but this behavior will be correct because the receiver doesn't know about ACK retransmission by the sender. 
This will be a momentary spike in the reading which will hopefully be correct when we enable continous rtt calculations for a flow.

This spike in the reading can be later correlated with tcp drops or something else as well.

bpf/rtt_tracker.h
dushyantbehl: Done

bpf/utils.h
dushyantbehl: done

bpf/flows.c
msherif1234: yeah I would move the SEC to each feature header even TC to the new hdr u added and just add a comment in the .c file stating this is a container file for all the different eBPF hooks the netobserv agent supports and we do so because bpf2go limit us to use only single `.c` file to generate go structs

bpf/maps_definition.h
msherif1234: no this is a run u can do urself on cluster-bot

bpf/rtt_tracker.h
msherif1234: ICMP support like what ePPing do if so that is limited to echo/echo reply so IMO its more of hack

bpf/rtt_tracker.h
msherif1234: ok for now pls doc this behavior so we have reference in case we need to revisit

pkg/agent/agent.go
dushyantbehl: Done.

pkg/agent/config.go
dushyantbehl: Fixed.

dushyantbehl:
> 

Fixed.

bpf/rtt_tracker.h
dushyantbehl: Added.

bpf/rtt_tracker.h
dushyantbehl: Yes that's what I meant for ICMP we will do for ECHO Request and ECHO reply id and seq. 
Can you please elaborate your concern with this as echo request/reply pair is something that is very widely used for latency measurement. We are just going to add support for getting the RTT value from request/reply pair and exporting it via the same flowlogs.

bpf/flows.c
dushyantbehl: Took care of the comment. Will be visible in the latest commit.

bpf/maps_definition.h
dushyantbehl: @msherif1234 Please check the PR description. Added the numbers

bpf/rtt_tracker.h
msherif1234: my point was echo isn't the only type for ICMP so claiming we do support for ICMP with only one type doesn't make much value IMO I did raised the same during ePPing meeting and they do agree

bpf/flows.c
```diff
@@ -13,38 +14,80 @@
...
+*/
 #include "tcp_drops.h"
+
+/* Defines a dns tracker,
+   which attaches at net_dev_queue hook. Is optional.
+*/
 #include "dns_tracker.h"
 
+/* Defines an rtt tracker,
+   which is run inside flow_monitor. Is optional.
```
msherif1234, line 33: nit: which runs

bpf/flows.c
```diff
@@ -13,38 +14,80 @@
...
+*/
+#include "rtt_tracker.h"
 
 static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
+
```
msherif1234, line 38: nit: white space

pkg/agent/agent.go
```diff
@@ -115,14 +115,24 @@ func FlowsAgent(cfg *Config) (*Flows, error) {
 		return nil, err
 	}
 
-	ingress, egress := flowDirections(cfg)
-
+	ingres, egress := flowDirections(cfg)
```
msherif1234, line 118: nit typo ingress

bpf/maps_definition.h
msherif1234: I will need clarifications on what u ran and associated the graphs I normally just label so what I expected main vs rtt off vs rtt on in the same graph for cpu and memory to make it easier to tell if there are any impacts if any

bpf/flows.c
dushyantbehl: Fixed.

bpf/flows.c
dushyantbehl: Fixed.

pkg/agent/agent.go
dushyantbehl: Fixed.

bpf/rtt_tracker.h
dushyantbehl: What I meant was not to call out full support for ICMP but rather that if we have the capability to RTT calculated from ICMP echo request/response pair.
But this is going tangential to the patch and is a point for future discussion I feel so maybe we can discuss this offline if you think.

msherif1234:
Thanks @dushyantbehl !!
/lgtm

bpf/maps_definition.h
dushyantbehl: @msherif1234 I noticed that the graphs were not tagged. Updated the PR description with new tagged graphs.

dushyantbehl: Resolving this. Thanks.

jotak:
/lgtm
/approve

@dushyantbehl thanks a lot and sorry for the long time to wait!

PR #118, opened 2023-05-05
WIP: change aggregation flow map to hashmap instead perCPU hashmap

msherif1234:

PR #119, opened 2023-05-10
NETOBSERV-975: avoid preallocating huge chunk of memory by default

msherif1234:
using BPF_F_NO_PREALLOC to allocate memory as needed and avoid default preallocation to help with better scale


msherif1234:
if we are migrating to global hashmap #118 then there is no need for this PR I am already taking care of the same there
/hold

dushyantbehl (approved):

dushyantbehl:
> if we are migrating to global hashmap #118 then there is no need for this PR I am already taking care of the same there /hold

@msherif1234 Even with a global hashmap I believe the spirit of your approach works which is just lazy allocation of the Hashmap memory so I feel this can be added even if we have global or local hashmap.

msherif1234:
> > if we are migrating to global hashmap #118 then there is no need for this PR I am already taking care of the same there /hold
> 
> @msherif1234 Even with a global hashmap I believe the spirit of your approach works which is just lazy allocation of the Hashmap memory so I feel this can be added even if we have global or local hashmap.

right the global hashmap PR has the same change as well hence we won't need this PR anymore once the #118 merge

PR #128, opened 2023-05-24
Fix dockerfile targetarch scope with docker

OlivierCazade:
Tested locally with both podman and docker.


jpinsonneau (approved):
Can't see the difference with my local versions but still works as expected :+1: 

Should we do that for all repos ?

jotak:
/lgtm
/approve
thanks!

PR #133, opened 2023-06-09
NETOBSERV-1091: remove CO-RE file and extensions as that causes douple allocations

msherif1234:
it was noticed cpu and memory are higher than 1.2 release 
that change was triggered by #110 
it seems using co-re results in a significant memory and cpu increase https://github.com/libbpf/libbpf/issues/697 is opened against libbpf to track this, issue has been seen even with latest version of the lib, we don't use CO-RE apis at all so for now we will remove it and once we have libbpf fix we will re-enable

diag on left side with this PR changes the diag on right is with 1.3 baseline
image
image

PR #135, opened 2023-06-15
WIP Fix memory and cpu scale issue work around in #133

msherif1234:
following up on https://github.com/cilium/ebpf/issues/1063 it seems we have a way to fix resources issues

also updated to go1.20 as well as linter version and that triggered few lint issues

this PR is blocking #115 the tracepoints hook doesn't work with disable CO-RE workaround


pkg/ebpf/tracer.go
```diff
@@ -81,7 +82,12 @@ func NewFlowFetcher(
 		}
 		return nil, fmt.Errorf("loading and assigning BPF objects: %w", err)
 	}
-
+	/*
+	 * since we load the program only when the we start we need to release
+	 * memory used by cached kernel BTF see https://github.com/cilium/ebpf/issues/1063
+	 * for more details.
+	 */
+	btf.FlushKernelSpec()
```
dushyantbehl, line 90: @msherif1234 do we also need to explicitly call the GC here after flushing kernel data structures as per your cilium cilium issue?

msherif1234: I am handling the flushing as well as adding gomemlimit in https://github.com/netobserv/netobserv-ebpf-agent/pull/140 maybe I should merge them in one PR

PR #140, opened 2023-06-21
NETOBSERV-1103: optimize ebpf agent map memory and cpu usage

msherif1234:
commit1:
- switch to use pointer to metric instead of metric
- manual trigger GC after flow eviction complete

commit2:
- remove CO-RE disable workaround added by #133 and `FlushKernelSpec`


msherif1234:
image (3)

pkg/flow/tracer_map.go
```diff
@@ -115,5 +116,6 @@ func (m *MapTracer) evictFlows(ctx context.Context, forwardFlows chan<- []*Recor
 	default:
 		forwardFlows <- forwardingFlows
 	}
+	runtime.GC() // Triggers a manual GC
```
jpinsonneau, line 119: I'm not sure about this manual call. For me the `GOMEMLIMIT` is supposed to be good enough if properly set.

Would it be interesting to make this configurable ? 
We can expose it in [debug](https://github.com/netobserv/network-observability-operator/blob/main/api/v1beta1/flowcollector_types.go#L222) section of the CRD.

msherif1234: manual call because under normal run not close to resource limit GOMEMLIMIT won't trigger GC so we will keep leaking flows map memory causing the memory to keep building up this manual call to prevent this memory buildup IMO

msherif1234:
image

jpinsonneau (dismissed):
LGTM in terms of code :+1: thanks @msherif1234 !

msherif1234:
Screenshot from 2023-06-22 16-52-45

pkg/agent/config.go
```diff
@@ -136,4 +136,6 @@ type Config struct {
 	KafkaSASLClientSecretPath string `env:"KAFKA_SASL_CLIENT_SECRET_PATH"`
 	// ProfilePort sets the listening port for Go's Pprof tool. If it is not set, profile is disabled
 	ProfilePort int `env:"PROFILE_PORT"`
+	// GoMemLimit sets soft memory cap to ebpf agent process
+	GoMemLimit string `env:"GOMEM_LIMIT"`
```
jotak, line 140: Why not using directly the GOMEMLIMIT env? That would avoid having to sync the env and the config. You would have it directly set by the operator.

pkg/flow/tracer_map.go
jotak: I agree with Julien, we know it has a CPU cost, having memory piling up is not bad per se as long as it's in the limit bounds configured.
GOMEMLIMIT should be sufficient to avoid OOM due to lack of gc calls, per my understanding.
But +1 to keep this option configurable so that we can still have a more aggressive GC strategy that we can turn on and off any time.

pkg/agent/config.go
msherif1234: I thought we wanted to control this value from operator ?

msherif1234: I get it operator will set the envvar nothing needed here I will remove this piece

pkg/flow/tracer_map.go
msherif1234: added envvar to control it and default it to true because we saw immediate memory gain with it at the cost of cpu pump I will explore options to better trim down those resources in the following PRs

jotak:
/lgtm
thanks!

PR #144, opened 2023-07-05
NETOBSERV-1129: Add kernel version chk to disable hooks not supported on older version

msherif1234:
TCPDrop hook doesn't exists on older kernelwe check for the following version and ignore tcpdrop for any older versions
```sh
sh-4.4# uname -r
5.14.0-284.20.1.el9_2.aarch64
```

- Testing with OCP 4.13 cluster
  - enable tcpDrop
```sh
$ oc get co
NAME                                       VERSION                              AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
authentication                             4.13.0-0.nightly-2023-06-20-224158   True        False         False      31m     

$ oc exec -i -n default dbgtools-deployment-85dff5ff6-92242 -- bpftool perf show
pid 46120  fd 8: prog_id 4  tracepoint  kfree_skb

```

- Testing with OCP 4.12 cluster
   - enable tcpdrop
```sh

$ oc get co
NAME                                       VERSION                              AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
authentication                             4.12.0-0.nightly-2023-07-03-233301   True        False         False      2m51s   
baremetal                                  4.12.0-0.nightly-2023-07-03-233301   True        False         False      18m     


$ oc logs -n netobserv-privileged netobserv-ebpf-agent-5cbmw 
time="2023-07-05T20:23:05Z" level=info msg="starting NetObserv eBPF Agent"
time="2023-07-05T20:23:05Z" level=info msg="initializing Flows agent" component=agent.Flows
time="2023-07-05T20:23:05Z" level=info msg="older kernel version not all hooks will be supported" component=utils
time="2023-07-05T20:23:06Z" level=info msg="push CTRL+C or send SIGTERM to interrupt execution"
time="2023-07-05T20:23:06Z" level=info msg="starting Flows agent" component=agent.Flows
time="2023-07-05T20:23:06Z" level=info msg="Flows agent successfully started" component=agent.Flows

$ oc exec -i -n default dbgtools-deployment-574c69cfb8-nnnfr -- bpftool perf show
$ 
```
- Added unit-test for the new api
- manual testing on OCP cluster
  - [X] OCP 4.10
  - [X] OCP 4.12
  - [X] OCP 4.13
  - [X] OCP 4.14


pkg/agent/agent.go
```diff
@@ -122,14 +122,15 @@ func FlowsAgent(cfg *Config) (*Flows, error) {
...
-		EnableRTT:     cfg.EnableRTT,
+		EnableIngress:    ingress,
+		EnableEgress:     egress,
+		Debug:            debug,
+		Sampling:         cfg.Sampling,
+		CacheMaxSize:     cfg.CacheMaxFlows,
+		TCPDrops:         cfg.EnableTCPDrops,
+		DNSTracker:       cfg.EnableDNSTracking,
+		EnableRTT:        cfg.EnableRTT,
+		SupportOldKernel: cfg.SupportOldKernel,
```
jotak, line 133: Actually, the agent could determine that on its own, without requiring a configuration?

jotak:
Code LGTM, maybe it can be improved by having the agent find the kernel version and determine itself if it must run without some hooks; this way we won't even have to worry about that on the operator side.

It must be tested with several ocp versions (ideally, all supported) - I can try some tomorrow...

pkg/agent/agent.go
msherif1234: kernel versions upstream and downstream don't align that will make it hard for me to check on special kernel

jotak: somewhere in comments you wrote: `For older kernel (< 5.17) kfree_sbk drop hook doesn't exists` so couldn't we just pick the kernel version and compare with 5.17 ?

pkg/ebpf/tracer.go
```diff
@@ -104,15 +108,51 @@ func NewFlowFetcher(cfg *FlowFetcherConfig) (*FlowFetcher, error) {
...
-		var ve *ebpf.VerifierError
-		if errors.As(err, &ve) {
-			// Using %+v will print the whole verifier error, not just the last
-			// few lines.
-			log.Infof("Verifier error: %+v", ve)
+	// For older kernel (< 5.17) kfree_sbk drop hook doesn't exists
+	if cfg.SupportOldKernel {
+		// Here we define another structure similar to the bpf2go created one but w/o the hooks that does not exist in older kernel
+		// Note: if new hooks are added in the future we need to update the following structures manually
+		type NewBpfPrograms struct {
```
dushyantbehl, line 115: This is a neat approach!! can be extended to various hooks and configurations like RTT also maybe!

msherif1234: atm there is only skb_free hook that isn't there in older kernel as we run into more cases we can generalize ?

dushyantbehl: yeah if this is only one hook.

dushyantbehl:
> Code LGTM, maybe it can be improved by having the agent find the kernel version and determine itself if it must run without some hooks; this way we won't even have to worry about that on the operator side.
> 
> It must be tested with several ocp versions (ideally, all supported) - I can try some tomorrow...

The userspace side of code can potentially run some command to find out kernel version if needed but would we wanna disable it at runtime or fail to run if older kernel is detected but config is set to enable all hooks. I am asking this because certain nodes of a cluster can have older kernel and in that case agent disabling some feature without informing might be a problem for some users.

msherif1234:
> > Code LGTM, maybe it can be improved by having the agent find the kernel version and determine itself if it must run without some hooks; this way we won't even have to worry about that on the operator side.
> > It must be tested with several ocp versions (ideally, all supported) - I can try some tomorrow...
> 
> The userspace side of code can potentially run some command to find out kernel version if needed but would we wanna disable it at runtime or fail to run if older kernel is detected but config is set to enable all hooks. I am asking this because certain nodes of a cluster can have older kernel and in that case agent disabling some feature without informing might be a problem for some users.

w/o doing anything the code as it stands will fail on older releases < 4.13 so the ask was to avoid failing on older release and I displayed warning plus in our doc we state clearly the kernel dependency so users will be aware

jotak:
Tested on 4.10 and works like a charm!

pkg/utils/utils.go
```diff
@@ -14,3 +22,70 @@ func GetSocket(hostIP string, hostPort int) string {
 	}
 	return socket
 }
+
+func IskernelOlderthan514() bool {
```
jotak, line 26: maybe a couple of unit tests for the new functions here, making sure it's bullet-proof? :-)

msherif1234: done

PR #146, opened 2023-07-06
NETOBSERV-588 Add IPv4/6 DSCP field to the exported flow

msherif1234:
tested with standalone flow collector for IPv4
```sh
 ping 1.1.1.1 -Q 0x60
PING 1.1.1.1 (1.1.1.1) 56(84) bytes of data.
64 bytes from 1.1.1.1: icmp_seq=1 ttl=53 time=18.4 ms
64 bytes from 1.1.1.1: icmp_seq=2 ttl=53 time=17.4 ms
64 bytes from 1.1.1.1: icmp_seq=3 ttl=53 time=17.9 ms
64 bytes from 1.1.1.1: icmp_seq=4 ttl=53 time=23.9 ms
64 bytes from 1.1.1.1: icmp_seq=5 ttl=53 time=15.3 ms
64 bytes from 1.1.1.1: icmp_seq=6 ttl=53 time=16.9 ms
64 bytes from 1.1.1.1: icmp_seq=7 ttl=53 time=18.7 ms
64 bytes from 1.1.1.1: icmp_seq=8 ttl=53 time=15.9 ms

ipv4: 09:45:36.374394 enp1s0 IP 192.168.122.69:0 > 1.1.1.1:0: dscp: 60 protocol:icmp type: 8 code: 0 dir:1 bytes:490 packets:5 flags:0 ends: 09:45:40.379900 dnsId: 0 dnsFlags: 0x0000 dnsReq: <nil> dnsRsp: <nil> rtt 0

```


bpf/types.h
```diff
@@ -77,6 +78,7 @@ typedef struct flow_metrics_t {
     // 0 otherwise
     // https://chromium.googlesource.com/chromiumos/docs/+/master/constants/errnos.md
     u8 errno;
+    u8 dscp;
```
dushyantbehl, line 81: @msherif1234 do you think with the growing structure we should turn this into a union of sub structures which cannot occur at the same time. 
Will save us space per entry of the hashmap as we grow to more protocols and types.

dushyantbehl: Maybe one 8 bit dhcp field is not too big a change to say this, I should have mentioned it before when tcp drops and dns was added but just mentioning this now to open disucssion.

PR #158, opened 2023-07-20
NETOBSERV-588 Add IPv4/6 DSCP field to the exported flow

msherif1234:
tested with standalone flow collector for IPv4
```sh
ping 1.1.1.1 -Q 0x60
PING 1.1.1.1 (1.1.1.1) 56(84) bytes of data.
64 bytes from 1.1.1.1: icmp_seq=1 ttl=53 time=18.4 ms
64 bytes from 1.1.1.1: icmp_seq=2 ttl=53 time=17.4 ms
64 bytes from 1.1.1.1: icmp_seq=3 ttl=53 time=17.9 ms
64 bytes from 1.1.1.1: icmp_seq=4 ttl=53 time=23.9 ms
64 bytes from 1.1.1.1: icmp_seq=5 ttl=53 time=15.3 ms
64 bytes from 1.1.1.1: icmp_seq=6 ttl=53 time=16.9 ms
64 bytes from 1.1.1.1: icmp_seq=7 ttl=53 time=18.7 ms
64 bytes from 1.1.1.1: icmp_seq=8 ttl=53 time=15.9 ms

ipv4: 09:45:36.374394 enp1s0 IP 192.168.122.69:0 > 1.1.1.1:0: dscp: 0x18 protocol:icmp type: 8 code: 0 dir:1 bytes:490 packets:5 flags:0 ends: 09:45:40.379900 dnsId: 0 dnsFlags: 0x0000 dnsReq: <nil> dnsRsp: <nil> rtt 0
```


bpf/flows.c
```diff
@@ -86,7 +86,7 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
             aggregate_flow->start_mono_time_ts = pkt.current_ts;
         }
         aggregate_flow->flags |= pkt.flags;
-
+        aggregate_flow->dscp = pkt.dscp;
```
jotak, line 89: not asking any change here, but just for curiosity, do you know if a single flow might have packets with different dscp values ?

jotak (approved):

bpf/flows.c
msherif1234: I was reading about this and it sounds the DSCP value is per flow this how they classify flow as higher pri over the others this is where I get that
https://datatracker.ietf.org/doc/html/rfc8837#name-dscp-mappings

msherif1234:
@Amoghrd when u can, can we give this is a go Thanks!!

PR #160, opened 2023-07-20
NETOBSERV-1198: drop_cause enum in vmlinux was generated on fc36 instead of rhel

msherif1234:
hook was checking for `SKB_DROP_REASON_NOT_SPECIFIED` value which supposed to be 2
based on https://elixir.bootlin.com/linux/v6.3/source/include/net/dropreason.h#L96
but the value in the current vmlinux is 1 so we were getting unclassified drops


jotak:
@msherif1234 how about this: https://github.com/netobserv/netobserv-ebpf-agent/blob/main/scripts/generators.Dockerfile#L1  ? shouldn't we change that as well?

msherif1234:
@jotak I am updating the Dockerfile to run some of the generators so we can build `.o` during our CI build but that PR isn't ready yet

PR #161, opened 2023-07-20
NETOBSERV-1198: drop_cause enum in vmlinux are based off old kernel

msherif1234:
hook was checking for SKB_DROP_REASON_NOT_SPECIFIED value which supposed to be 2
based on https://elixir.bootlin.com/linux/v6.3/source/include/net/dropreason.h#L96
but the value in the current vmlinux is 1 so we were getting unclassified drops
I updated drop_reason_enum based on kernel that supported this hook `rhel9`


jotak:
> I generated vmlinux off rhel9 OCP node

@msherif1234 is that going to be incompatible with rhel8-based ocp? @Amoghrd perhaps it should be tested on ocp4.12 as well?

msherif1234:
tcpdrop is disabled with older kernels

Amoghrd:
@jotak Yes, could confirm that the eBPF pod comes up fine with older versions of OCP(tested with 4.12) and could view the log line `level=info msg="older kernel version not all hooks will be supported" component=utils`.

bpf/headers/vmlinux.h
```diff
@@ -35484,71 +35484,75 @@ struct phc_vclocks_reply_data {
...
+	SKB_DROP_REASON_TAP_TXFILTER = 60,
+	SKB_DROP_REASON_ICMP_CSUM = 61,
+	SKB_DROP_REASON_INVALID_PROTO = 62,
+	SKB_DROP_REASON_IP_INADDRERRORS = 63,
+	SKB_DROP_REASON_IP_INNOROUTES = 64,
+	SKB_DROP_REASON_PKT_TOO_BIG = 65,
+	SKB_DROP_REASON_DUP_FRAG = 66,
+	SKB_DROP_REASON_FRAG_REASM_TIMEOUT = 67,
+	SKB_DROP_REASON_FRAG_TOO_FAR = 68,
+	SKB_DROP_REASON_MAX = 69,
```
jpinsonneau, line 35555: I'm not confortable at all with that... How can we ensure we are up to date on this enum ?

jpinsonneau, line 35555: If I'm not wrong:

```suggestion
	SKB_DROP_REASON_FRAG_TOO_FAR = 68,
	SKB_DROP_REASON_TCP_MINTTL = 69,
	SKB_DROP_REASON_IPV6_BAD_EXTHDR = 70,
	SKB_DROP_REASON_IPV6_NDISC_FRAG = 71,
	SKB_DROP_REASON_IPV6_NDISC_HOP_LIMIT = 72,
	SKB_DROP_REASON_IPV6_NDISC_BAD_CODE = 73,
	SKB_DROP_REASON_IPV6_NDISC_BAD_OPTIONS = 74,
	SKB_DROP_REASON_IPV6_NDISC_NS_OTHERHOST = 75,
	SKB_DROP_REASON_MAX = 76,
	SKB_DROP_REASON_MAX = 77,
```

jpinsonneau: are we forced to mention values for each enum ?

msherif1234: this is kernel definition I checked with rhel and they normally append to the end with new causes except for this https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=0e84afe8ebfbb9eade3f4f6de4720887bf908e26 which cause teh disconnect in general I will be watching this vmlinux and make sure updates are compatible with older releases even if its at the cost of maintaining custom version for it

msherif1234: yeah I can add those too Thanks!!

jpinsonneau (approved):
LGTM, thanks @msherif1234

PR #162, opened 2023-07-21
Avoid using LRU hashmap because they consume lots of memory

msherif1234:
Even when the feature is not on memory with latest showing double the normal use

Note: In general relying on LRU to evict old entries isn;t recommended because its hard to perdict when an entry will be deleted
at somepoint we need to add eviction logic in userspace

latest_perf


dushyantbehl (approved):
Makes sense to do. LRU is taking too much memory. 
Userspace eviction can be added later for stale entries.

PR #163, opened 2023-07-22
Evict DNS and RTT stale entries everytime we evict flows table

msherif1234:
Its possible we can have DNS queries w/o any responses or RTT entries w/o replies those entries will be there for ever, since we can't use LRU because we don't know when entries will be deleted and its very large memory footprint we will evict those entries in the user space every time we evict flows hash table by default every 5 seconds


dushyantbehl:
@msherif1234 can we add a check here like clear the entries which are 5s/10s earlier and are still there (which means they are stale I mean)?
we need to check this against Mono clock.

msherif1234:
> @msherif1234 can we add a check here like clear the entries which are 5s/10s earlier and are still there (which means they are stale I mean)? we need to check this against Mono clock.

@dushyantbehl I am not following, this code will be invoked every 5s so entries at this points are been there for 5s w/o been deleted by bpf code, can u explain/clarify more ?

dushyantbehl:
Thanks @msherif1234 for adding the timeout. I just feel one small thing that since we are handling three maps with one funciton maybe we can make a map manager and handle all three inside of it..
Currently the flow is passed via `LookupAndDeleteMap` function which acts on bpf metrics and is a bit misleading.

dushyantbehl:
other than that looks good to me

msherif1234:
> Thanks @msherif1234 for adding the timeout. I just feel one small thing that since we are handling three maps with one funciton maybe we can make a map manager and handle all three inside of it.. Currently the flow is passed via `LookupAndDeleteMap` function which acts on bpf metrics and is a bit misleading. WDYT?
> 
> other than that looks good to me.

I created new interface to handle features map I think is more cleaner WDYT ?

jotak (approved):

PR #167, opened 2023-07-31
NETOBSERV-1223: check Maps to make sure not nil b4 iterating

msherif1234:
there is some behavior difference with maps between rhel8 and rhel9
on rhel8 it seems maps not even allocated until its populated leading to this issue while in rhel9 that doesn't seem to be the case.

To be on safe side lets make sure maps pointer is != nil before doing invoking `Iterate`


jpinsonneau (approved):
:+1:

PR #172, opened 2023-08-29
NETOBSERV-1268: handle concurrency issues between kernel and userspace

msherif1234:
with using global hmap we increased the window where both userspace and kernel can collide to we can't use`bpf_spin_lock()` because tracepoints doesn't allow it so we have to revert back to use perCPU map :(

testing
- verified unaccuracy is now more accurate with sampling of 1
- verified PktDrop still working
- Verified DNS tracking
- measure resources impact going back to perCPU hmap
Signed-off-by: msherif1234 <mmahmoud@redhat.com>


jotak:
I think this confirms my doubts, the problem isn't solved: I still get low bytes counter on the chunks, except for the last one which is more consistent: : 
image
(this is for a 300MB download, you can see only 60MB is captured)

This totally makes sense if there is a concurrency issue between cores:
- During downloads, several cores are updating the map concurrently, each overwriting what the other captures
- When all cores but one have finished the processing, concurrency issues disappear hence everything is correctly reported, that's why the last chunk is always bigger.

jotak:
To test:
- `oc get pods -n netobserv` and choose a FLP pod (or any pod that has curl)
- run `oc exec -it <pod that has curl> -- curl https://archives.fedoraproject.org/pub/archive/fedora/linux/releases/30/Cloud/x86_64/images/Fedora-Cloud-Base-Vagrant-30-1.2.x86_64.vagrant-libvirt.box  --output /tmp/test` (this is a random 300MB image)
- open console at https://<your cluster>/netflow-traffic?timeRange=300&limit=50&match=all&showDup=false&function=last&type=bytes&packetLoss=all&recordType=flowLog&filters=src_owner_name%3D%22%22%3Bdst_namespace%3D%22netobserv%22&bnf=false

You should see the flows similar to my screenshot above.
If you have all in one flow (no chunk), try with a bigger image, or maybe try `--rate-limit` with curl.

bpf/flows.c
```diff
@@ -81,16 +80,7 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
...
-        long ret = bpf_map_update_elem(&aggregated_flows, &id, aggregate_flow, BPF_ANY);
-        if (trace_messages && ret != 0) {
-            // usually error -16 (-EBUSY) is printed here.
-            // In this case, the flow is dropped, as submitting it to the ringbuffer would cause
-            // a duplicated UNION of flows (two different flows with partial aggregation of the same packets),
-            // which can't be deduplicated.
-            // other possible values https://chromium.googlesource.com/chromiumos/docs/+/master/constants/errnos.md
-            bpf_printk("error updating flow %d\n", ret);
-        }
+        bpf_spin_unlock(&aggregate_flow->lock);
```
msherif1234, line 83: this is update in place we get pointer to flow in memory so we can update it w/o need to call `bpf_map_update_elem()` like this example https://elixir.bootlin.com/linux/latest/source/tools/testing/selftests/bpf/progs/test_map_lock.c

jotak:
fyi I forgot to mention in my test steps above: `Install netobserv/FlowCollector with sampling=1` (above comment updated)

msherif1234:
cc'd @tohojo as discussed offline we have no way to ensure concurrency using global hmap now its shared with tracepoints hook so we have to fall back to use perCPU hmap

msherif1234:
image

msherif1234:
revert_hmap
as expected there are slight increase in CPU and memory going back to perCPU hmap

dushyantbehl (commented):
Have not tested myself but changes look fine to me and seems like a revert to the old code we had. Are we doing anything different than that @msherif1234 ?

msherif1234:
> Have not tested myself but changes look fine to me and seems like a revert to the old code we had. Are we doing anything different than that @msherif1234 ?

one diff from the old code is in lookupanddelete it used to return []metrics now it return []*metrics , also the aggregate code has now the new features to aggregate

jpinsonneau:
I'm also seeing issues on bytes count. Steps to reproduce:
- set sampling 1
- do a curl between two httpd pods using:
```
sh-4.4$ curl --compressed -so /dev/null 10.129.0.36:8080 -w '%{size_download}'
4927 // this is the expected size
```
- see the amount of bytes in flow table

Using this PR I get `140` whereas without it I had `5485`

pkg/flow/tracer_map.go
```diff
@@ -126,3 +126,23 @@ func (m *MapTracer) evictFlows(ctx context.Context, enableGC bool, forwardFlows
...
+		mtlog.Warn("invoked aggregate with no values")
+		return &ebpf.BpfFlowMetrics{}
+	}
+	aggr := &ebpf.BpfFlowMetrics{}
+	for _, mt := range metrics {
+		if mt != nil {
+			// eBPF hashmap values are not zeroed when the entry is removed. That causes that we
+			// might receive entries from previous collect-eviction timeslots.
+			// We need to check the flow time and discard old flows.
+			if mt.StartMonoTimeTs <= m.lastEvictionNs || mt.EndMonoTimeTs <= m.lastEvictionNs {
```
jpinsonneau, line 141:
```suggestion
			if mt.EndMonoTimeTs <= m.lastEvictionNs {
```

Can you please explain why we need to compare Start and End times here ?

It seems both curl on httpd page & ubuntu image download works fine when removing the compare on start :smile_cat:

jpinsonneau: image
image

msherif1234: > Can you please explain why we need to compare Start and End times here ?
> 
> It seems both curl on httpd page & ubuntu image download works fine when removing the compare on start 😸

that code was just a revert and that is what 1.3 has where u don't see the issue correct ?

bpf/utils.h
```diff
@@ -253,12 +253,13 @@ static inline long pkt_drop_lookup_and_update_flow(struct sk_buff *skb, flow_id
                                                    enum skb_drop_reason reason) {
      flow_metrics *aggregate_flow = bpf_map_lookup_elem(&aggregated_flows, id);
      if (aggregate_flow != NULL) {
+         aggregate_flow->end_mono_time_ts = bpf_ktime_get_ns();
```
jotak, line 256: Previously, there was some specific handling of empty start time here:
```
        // it might happen that start_mono_time hasn't been set due to
        // the way percpu hashmap deal with concurrent map entries
        if (aggregate_flow->start_mono_time_ts == 0) {
            aggregate_flow->start_mono_time_ts = current_time;
        }
```

cf also https://github.com/netobserv/netobserv-ebpf-agent/commit/c696bb0b7e4243845fe5289ac4a07727df517401

But I admit I don't really understand this old patch... I don't see in which cases `start_mono_time_ts` could not be set.

pkg/agent/agent_test.go
```diff
@@ -183,10 +219,13 @@ func testAgent(t *testing.T, cfg *Config) *test.ExporterFake {
...
-	key1Metrics := ebpf.BpfFlowMetrics{Packets: 3, Bytes: 44, StartMonoTimeTs: now + 1000, EndMonoTimeTs: now + 1_000_000_000}
+	key1Metrics := []ebpf.BpfFlowMetrics{
+		{Packets: 3, Bytes: 44, StartMonoTimeTs: now + 1000, EndMonoTimeTs: now + 1_000_000_000},
+		{Packets: 1, Bytes: 22, StartMonoTimeTs: now, EndMonoTimeTs: now + 3000},
+	}
 
-	ebpfTracer.AppendLookupResults(map[ebpf.BpfFlowId]*ebpf.BpfFlowMetrics{
-		key1: &key1Metrics,
+	ebpfTracer.AppendLookupResults(map[ebpf.BpfFlowId][]ebpf.BpfFlowMetrics{
+		key1: key1Metrics,
```
jotak, line 228: This isn't testing the dedup, we should restore this test as it was previously:

```
	key1Metrics := []ebpf.BpfFlowMetrics{
		{Packets: 3, Bytes: 44, StartMonoTimeTs: now + 1000, EndMonoTimeTs: now + 1_000_000_000},
		{Packets: 1, Bytes: 22, StartMonoTimeTs: now, EndMonoTimeTs: now + 3000},
	}
	key2Metrics := []ebpf.BpfFlowMetrics{
		{Packets: 7, Bytes: 33, StartMonoTimeTs: now, EndMonoTimeTs: now + 2_000_000_000},
	}

	ebpfTracer.AppendLookupResults(map[ebpf.BpfFlowId][]ebpf.BpfFlowMetrics{
		key1:     key1Metrics,
		key1Dupe: key1Metrics,
		key2:     key2Metrics,
	})
```

.. and adapt the expectations in the related tests (dedup test expects 2 results, dedup-just-mark test expects 3 results with one flagged duplicate, no-dedup expects 3 results)

pkg/flow/tracer_map.go
jotak: @msherif1234 could it be due to this code block being removed: (as said here https://github.com/netobserv/netobserv-ebpf-agent/pull/172#discussion_r1316971958  ) ?
```
        // it might happen that start_mono_time hasn't been set due to
        // the way percpu hashmap deal with concurrent map entries
        if (aggregate_flow->start_mono_time_ts == 0) {
            aggregate_flow->start_mono_time_ts = current_time;
        }
```

It seems like sometimes `start_mono_time_ts` is 0, which would indeed make the condition `mt.StartMonoTimeTs <= m.lastEvictionNs` be true ...

bpf/utils.h
msherif1234: @Joel this drop handling function

msherif1234: but u are right that part is missing in flows.c the revert didn't bring it back

pkg/flow/tracer_map.go
msherif1234: yeah that is the proof that same flow won't always hit the same cpu core so we had to add this trick unfortuntely revert couldn't bring that code and I forget about it thanks for catching it now I tested both large and small curl and they are pretty accurate

pkg/agent/agent_test.go
msherif1234: done

jotak:
not tested - code-wise this looks good to me, now we must check if @jpinsonneau 's test is satisfied

msherif1234:
I did number of small curls b2b and they are consistent 
image

jpinsonneau:
Yeah I confirm both small curl / 2GB download scenarios works fine :+1: 
Thanks guys !

msherif1234:
@memodi @Amoghrd can u pls sanity check this PR and make sure the new features still working @dushyantbehl can u pls check RTT functionality with this PR

Amoghrd:
Sanity checked PacketDrop with this PR and everything looks good; works as expected!

PR #173, opened 2023-09-04
NETOBSERV-1240 DNS flows can be missed with Duplicates "unselected" by default in UI

jpinsonneau:
This PR skip `dedupe` on DNS flows.

To avoid having the request and response showing the same Direction when service is involved, we should fix the related case:
https://github.com/netobserv/flowlogs-pipeline/pull/483#issuecomment-1705332173


jotak:
it doesn't looks good, as we would end up with multiple duplicates showing "Duplicate=false"

jpinsonneau:
Closing this in favor of alternative PR

PR #174, opened 2023-09-07
NETOBSERV-1240: prefer flows with DNS records when checking for dup

msherif1234:


pkg/flow/deduper.go
```diff
@@ -26,6 +26,7 @@ type deduperCache struct {
 
 type entry struct {
 	key        *ebpf.BpfFlowId
+	record     *Record
```
jotak, line 29: I was thinking about this solution to store the whole record here but it has some downsides / challenges:

- Records are now kept in memory in two places: in the `Accounter` and now here - with different eviction timeouts, the one defined here being bigger for good reasons. So we can expect an increased memory usage as those records won't be garbage-collected until they expire from this cache.
- If a record was flushed out from `Accounter` and still present here while you revert its Duplicate flag with `fEntry.record.Duplicate = true`, this will actually have no effect downstream, so there are still cases where we'll have duplicates not flagged as such

jotak: I wonder, does the concept of "Weak reference" exists in go? In order to keep this cache with weak refs of records, so that they're still gc'ed after Accounter flush ?

pkg/flow/deduper.go
```diff
@@ -64,8 +58,8 @@ func Dedupe(expireTime time.Duration, justMark bool) func(in <-chan []*Record, o
 
 // isDupe returns whether the passed record has been already checked for duplicate for
 // another interface
-func (c *deduperCache) isDupe(key *ebpf.BpfFlowId) bool {
-	rk := *key
+func (c *deduperCache) isDupe(r *Record, justMark bool, fwd *[]*Record) {
```
jotak, line 61: this function should be renamed as it doesn't return bool anymore, maybe something like `checkDupe` ?

pkg/flow/deduper.go
```diff
@@ -79,17 +73,36 @@ func (c *deduperCache) isDupe(key *ebpf.BpfFlowId) bool {
...
 		// The input flow is duplicate if its interface is different to the interface
 		// of the non-duplicate flow that was first registered in the cache
-		return fEntry.ifIndex != key.IfIndex
+		// except if the new flow has DNS enrichment in this case we will
+		// mark the original flow as duplicate
+		if r.DNSLatency != 0 && fEntry.record.DNSLatency == 0 && fEntry.ifIndex != r.Id.IfIndex {
+			if justMark {
+				fEntry.record.Duplicate = true
+			} else {
+				fwd = findAndDeleteRecord(fwd, fEntry.record)
```
jotak, line 82: This won't work if the entry was in cache but already sent previously (via a previous set of records): it won't be in the `fwd` slice.

pkg/flow/deduper.go
msherif1234: renamed

msherif1234:
dup

pkg/flow/deduper.go
msherif1234: so we are storing pointer to the record not the record itself I shared the memory usage below b4 and after and its about the same

pkg/flow/deduper.go
msherif1234: based on the info in the arch doc my understanding is the dup stage after the accout , plus of the record not in the same fwd slice in that case it won't be considered dup right ?

jotak: I'm wrong about the accounter but IMO the problem is still the same: a flow could have been sent out to FLP and its dup comes in later.

> the record not in the same fwd slice in that case it won't be considered dup right ?

I think that's wrong and that's why there is a separate cache with its own expiry time: we keep track of flow keys in that cache, even for flows already sent to FLP, so that flows coming later can still be seen as duplicates

msherif1234:
image

image

msherif1234:
@memodi can u pls give this a try and see if it fixes the issue ?

memodi:
Verified DNS flows shows up without selecting "Show Duplicates" in UI - cc @skrthomas 
/label qe-approved

jotak (approved):
lgtm!
thanks!

PR #206, opened 2023-10-12
NETOBSERV-1245: NETOBSERV-1304: fix TCP DNS query

msherif1234:
## Description

Tracking DNS over TCP was impossible using tracepoints as the dns hdr never exists in the 1st segment and when use ebpf TP there are no helpers to get to packet from the following segments

Solution was to do DNS tracking inline with TC hook and to gain better accuracy for all NDS queries its needed to run with sampling of 1
image

image

## Dependencies

n/a

## Checklist

* [X] Does this PR require product documentation?
* [] Does this PR require a product release notes entry?


msherif1234:
image

jpinsonneau (approved):
That's awesome ! thanks @msherif1234 

Do you expect better performances since the DNS tracepoint is not used anymore ? 
As followup, should we have a separated `sampling` per feature (DNS / RTT) ? Or is there a way to avoid having sampling `1` to make it accurate.

msherif1234:
cc'd @Amoghrd

msherif1234:
> That's awesome ! thanks @msherif1234
> 
> Do you expect better performances since the DNS tracepoint is not used anymore ? As followup, should we have a separated `sampling` per feature (DNS / RTT) ? Or is there a way to avoid having sampling `1` to make it accurate.

its not only for DNS/RTT even for the regular TC flows if sampling not 1 won't get very accurate latency as well, I spoke with @jotak and he told me for customers who are looking for accurate analysis its expected/encouraged to use sampling of `1`

PR #217, opened 2023-10-25
NETOBSERV-1298: include duplicate list in the exported record

msherif1234:
## Description

create duplicate list which contains interface name and direction pairs so we can
avoid sending duplicate flows
## Dependencies

n/a

## Checklist


jotak (commented):
Thanks @msherif1234 , sounds good to me on the deduper side!
Then the protobuf will have to be updated accordingly

pkg/decode/decode_protobuf_test.go
```diff
@@ -60,6 +66,7 @@ func TestDecodeProtobuf(t *testing.T) {
 			},
 			expected: &config.GenericMap{
 				"FlowDirection":   1,
+				"FlowDirections":  []interface{}{1},
```
jotak, line 69: We need to discuss about what's the best model here to have, regarding the loki querying side too 

At the moment I don't think our json maps have other value types than strings or numbers, so arrays would be something new here.

In the JIRA description I was suggesting a single field, such as "TraversedInterfaces" for both interface+direction, in the form of ["genev/0", "abc1234/1"], but this isn't perhaps a better solution and is still an array.

Another option could be coma-separated strings instead of arrays, which perhaps would be easier to manipulate in queries with regexp.

@jpinsonneau any thoughts here?

jotak: Also, to avoid ambiguity and redundancy, I think we could remove FlowDirection & Interface (singular) when the plural form is defined? (and vice-versa)

jpinsonneau: As soon as we will not need the deduplication mechanism on loki side anymore, we can get rid of `FlowDirection` label and `Interface` field indeed.

What would be the issue with JSON arrays for each of these then ?
If the JSON value is:
```
{
  "FlowDirections": [0,1],
  "Interfaces": ["eth0","br-ex"],
  ...
}
```

Regexes could be the following:
```
"Interfaces": \[.*"br-ex".*\]
```
to match only `br-ex` interface.

```
"FlowDirections": \[.*0.*\]
```
to match only ingress.

I don't think we need to cover exact match here since there is no overlap between Interfaces names or FlowDirection numbers :thinking:

jotak: I don't know what the problem could be with arrays, I'm asking because I guess new implementation is needed with the configurable fields both for filtering and for column display, whereas I guess we already have this in place for strings

jpinsonneau: Yes there is a bit more to manage on query side but that's fine to me.
Internally we already bind arrays to column [value function](https://github.com/netobserv/network-observability-console-plugin/blob/main/web/src/utils/columns.ts#L101) so I'm confident on the web side.

jotak: ok ... anyway we can still change that later if we find this isn't ideal on consuming side

PR #218, opened 2023-10-26
NETOBSERV-1379: enhance DNS debugging to dbg DNS over TCP with NA fields

msherif1234:
## Description

it was noticed with DNS over TCP that the handshake packets are too small and it doesn't contain any DNS data from UI DNS fields showed as `n/a` which is true

this PR propagate the DNS lookup return code so we can differentiate between cases like TCP handshake case and real errors enriching with DNS info.

when we get N/A `bpf_skb_load_bytes()` return an error
```golang
working
======
 "dns_record": {
                        "id": 40514,
                        "flags": 34176,
                        "latency": 185794,
                        "errno": 0,
                        "offset": 68,
                        "tcp_len": 34,
                        "skb_len": 291
                    },

none working
==========
 "dns_record": {
                        "id": 0,
                        "flags": 0,
                        "latency": 0,
                        "errno": 7,
                        "offset": 68,
                        "tcp_len": 34,
                        "skb_len": 66
                    },

```
in the none working case the received packet is very small that is why we can't fetch DNS header info.

Those small packets are TCP handshake and those packets don't have data by definition this clear when u look at the pcap

image (1)

## Dependencies

https://github.com/netobserv/network-observability-console-plugin/pull/425
https://github.com/netobserv/flowlogs-pipeline/pull/533

## Checklist

* [x] Does this PR require product documentation?
  * [x] If so, make sure the JIRA epic is labelled with "documentation" and provides a description relevant for doc writers, such as use cases or scenarios. Any required step to activate or configure the feature should be documented there, such as new CRD knobs.


msherif1234:
when we get N/A bpf_skb_load_bytes() return an error
```sh
Not working
==========
"dns_record": {
                        "id": 0,
                        "flags": 0,
                        "latency": 0,
                        "errno": 14 <<<<<<<<<<<<<<< "EFAULT"
                    },

working
=======
dns_record": {
                        "id": 31706,
                        "flags": 34176,
                        "latency": 228919,
                        "errno": 0
                    },
```

jpinsonneau:
@msherif1234 are there error codes we should expect ?
https://elixir.bootlin.com/linux/v4.7/source/include/uapi/asm-generic/errno-base.h#L10

msherif1234:
> @msherif1234 are these error codes we should expect ? https://elixir.bootlin.com/linux/v4.7/source/include/uapi/asm-generic/errno-base.h#L10

Correct

pkg/decode/decode_protobuf.go
```diff
@@ -90,11 +90,13 @@ func PBFlowToMap(flow *pbflow.Record) config.GenericMap {
 		}
 	}
 
+	out["DnsErrno"] = flow.GetDnsErrno()
```
jpinsonneau, line 93: As discussed, my concern here is adding unecessary field on non dns related flows.
We agreed to keep it as is for now.

jpinsonneau (approved):
LGTM, just a note for future improvment.
Thanks @msherif1234 !

PR #223, opened 2023-11-16
Add a new exporter type: DirectFLP

jotak:
## Description

DirectFLP starts the whole flowlogs-pipeline process directly from the
Agent, eliminating the need for a network connection between them.

This can be useful when both Agent and FLP run on the same host, saving
resources. This is mutually exclusive with other exporters such as with
Kafka.

## Dependencies

https://github.com/netobserv/flowlogs-pipeline/pull/538

(hold until this is merged => replace directive to remove)

## Checklist

  * [x] No QE (e.g. trivial change with high reviewer's confidence, or per agreement with the QE team).


pkg/agent/config.go
```diff
@@ -140,6 +140,9 @@ type Config struct {
 	KafkaSASLClientSecretPath string `env:"KAFKA_SASL_CLIENT_SECRET_PATH"`
 	// ProfilePort sets the listening port for Go's Pprof tool. If it is not set, profile is disabled
 	ProfilePort int `env:"PROFILE_PORT"`
+	// Flowlogs-pipeline configuration, used when export is "direct-flp". Cf https://github.com/netobserv/flowlogs-pipeline
+	// The "ingest" stage must be omitted from this configuration, since it is handled internally by the agent.
+	FLPConfig string `env:"FLP_CONFIG"`
```
jpinsonneau, line 145: As soon as `FLP_CONFIG` is set, `EXPORT` should be ignored.
The export capabilities from pod point of view are inside FLP config right ?

pkg/agent/agent.go
```diff
@@ -251,6 +251,8 @@ func buildFlowExporter(cfg *Config) (node.TerminalFunc[[]*flow.Record], error) {
 		return buildIPFIXExporter(cfg, "udp")
 	case "ipfix+tcp":
 		return buildIPFIXExporter(cfg, "tcp")
+	case "direct-flp":
+		return buildDirectFLPExporter(cfg)
```
jpinsonneau, line 255: I would move that outside of cfg.Export as:

```
if len(cfg.FLPConfig) {
  return buildDirectFLPExporter(cfg)
}
switch cfg.Export {
  ...
}
```

jpinsonneau (commented):
Approach sounds good ! Just a small change on config to make it clear for the user
We should also add an example in `deployments` folder

pkg/agent/config.go
jotak: I guess it depends from where you sit.. In my mind, this new mode is really a new exporter, sitting from the Agent codebase PoV: the agent generates flows and then exports them somewhere, here to FLP; the fact that FLP is in the same process is a detail, it's still an exporter from this codebase PoV.

I guess you have more the "external user point of view" where, once FLP config is set, "export" would mean what happens "after FLP processing", which indeed doesn't really make sense since FLP does the export (to loki, prom, etc.)

idk, I think both approaches can be defended ... @msherif1234 , some preference?

msherif1234: thinking of it as new exporter mode resonate well with me as well

pkg/exporter/direct_flp.go
```diff
@@ -0,0 +1,44 @@
...
+	*ingest.InProcess
+}
+
+func StartDirectFLP(jsonConfig string) (*DirectFLP, error) {
+	var cfg flpconfig.ConfigFileStruct
+	// Note that, despite jsonConfig being json, we use yaml unmarshaler because the json one
+	// is screwed up for HTTPClientConfig in github.com/prometheus/common/config (used for Loki)
+	// This is ok as YAML is a superset of JSON.
+	// E.g. try unmarshaling `{"clientConfig":{"authorization":{}}}` as a api.WriteLoki
+	// See also https://github.com/prometheus/prometheus/issues/11816
```
msherif1234, line 24: interesting thanks for referring to the issue in promo its almost 1yr so probably this work around will stay for sometime

pkg/exporter/direct_flp.go
```diff
@@ -0,0 +1,44 @@
...
+
+	return &DirectFLP{InProcess: flp}, nil
+}
+
+// ExportFlows accepts slices of *flow.Record by its input channel, converts them
+// to *pbflow.Records instances, and submits them to the collector.
+func (d *DirectFLP) ExportFlows(input <-chan []*flow.Record) {
+	for inputRecords := range input {
+		pbRecords := flowsToPBNoChunk(inputRecords)
+		d.Write(pbRecords)
```
msherif1234, line 42: This step can be removed if we let the new FLP pipe process ebpf record directly not protobuf records

pkg/exporter/direct_flp.go
jotak: Yeah there's even a PR from @OlivierCazade that is older than that :-) https://github.com/prometheus/common/pull/394

pkg/exporter/direct_flp.go
jotak: done via [bcccbb7](https://github.com/netobserv/netobserv-ebpf-agent/pull/223/commits/bcccbb7dfca1b8ec9f1e8940911dce6c59376660)

pkg/decode/decode_protobuf.go
```diff
@@ -67,26 +67,30 @@ func PBFlowToMap(flow *pbflow.Record) config.GenericMap {
...
-		if proto == syscall.IPPROTO_TCP || proto == syscall.IPPROTO_UDP || proto == syscall.IPPROTO_SCTP {
-			if proto == syscall.IPPROTO_TCP {
-				out["SrcPort"] = flow.Transport.GetSrcPort()
-				out["DstPort"] = flow.Transport.GetDstPort()
-				out["Flags"] = flow.Flags
-			} else {
-				out["SrcPort"] = flow.Transport.GetSrcPort()
-				out["DstPort"] = flow.Transport.GetDstPort()
-			}
+	if ethType != ethernet.EtherTypeIPv4 && ethType != ethernet.EtherTypeIPv6 {
```
jotak, line 71: @msherif1234 I'm doing this early return here when it's not ipv4/ipv6, could you confirm this is correct?
Previously, in that case (e.g. with ARP), the code would still check for drops/DNS/RTT fields and now this isn't the case anymore, but I believe our drops/DNS/RTT metrics wouldn't make sense for non-IP packets, right?

If this doesn't sound good to you, I can rollback, but I'd need to adapt also the ConvertToFLP function because I'm enforcing in my tests that the same flow input produces the exact same output regardless if it's using ConvertToFLP or converting via protobuf.

pkg/exporter/convert_flp_test.go
```diff
@@ -0,0 +1,376 @@
...
+	"google.golang.org/protobuf/proto"
+)
+
+func TestConvertFLP(t *testing.T) {
+	decoder := decode.Protobuf{}
+
+	someTime := time.Now()
+	var someDuration time.Duration = 10000000 // 10ms
+
+	tests := []struct {
```
jotak, line 24: @msherif1234 FYI the tests here are the same that were previously in `decode_protobuf_test` except that they're now used to test both pb conversion and direct FLP conversion, making sure they both produce the same result

pkg/decode/decode_protobuf.go
msherif1234: the only thing we could be missing here is l2 drops, also pls test with DNS TCP to be sure `DNSError` is only none zero for TCP handshake I can help you with this test if u like

pkg/decode/decode_protobuf_test.go
```diff
@@ -7,375 +7,11 @@ import (
...
-		rawPB, err := proto.Marshal(tt.flow)
-		require.NoError(t, err)
-		out, err := decoder.Decode(rawPB)
-		require.NoError(t, err)
-		assert.NotZero(t, out["TimeReceived"])
-		delete(out, "TimeReceived")
-		assert.Equalf(t, *tt.expected, out, tt.name)
-	}
-}
+// NOTE: more tests in convert_flp_test.go
```
msherif1234, line 14: did intentionally deleted conversion from gRPC to generic map test cases here ?

pkg/exporter/convert_flp_test.go
msherif1234: this test coverting from ebpf record to flp configmap but it seems u deleted from gRPC to configmap in case this mode isn't set ?

pkg/decode/decode_protobuf_test.go
jotak: yes, like I wrote this test was moved (and sightly changed) to convert_flp_test, but I admit the name is misleading, I'll rename that file to clarify it's testing BOTH grpc and direct-flp conversions

pkg/exporter/convert_flp_test.go
jotak: cf my next comment below

pkg/exporter/convert_flp_test.go
```diff
@@ -0,0 +1,376 @@
...
+	for _, tt := range tests {
+		// Test direct conversion
+		outDirect := ConvertToFLP(tt.flow)
+		assert.NotZero(t, outDirect["TimeReceived"])
+		delete(outDirect, "TimeReceived")
+		err := normalizeMap(outDirect)
+		require.NoError(t, err)
+		assert.Equalf(t, *tt.expected, outDirect, tt.name)
+
+		// Test pb conversion
```
jotak, line 348: see here @msherif1234 : I am testing both FLP conversion AND protobuf conversion, with the bonus thatit makes sure the two are producing the same result (they're both compared against the same expected map)

jotak:
thanks @msherif1234 !
/approve

PR #245, opened 2024-01-03
WIP: NETOBSERV-1390: Flow RTT values are n/a

msherif1234:
## Description

using tcp_rcv kprobe to reuse kernel calculated rtt directly from TCP socket instead of trying to calculate it
## Dependencies

n/a

## Checklist


msherif1234:
main vs this PR both with RTT enabled

image

bpf/rtt_tracker.h
```diff
@@ -1,143 +1,161 @@
...
+}
+
+static inline int rtt_lookup_and_update_flow(flow_id *id, u16 flags, u64 rtt) {
+    flow_metrics *aggregate_flow = bpf_map_lookup_elem(&aggregated_flows, id);
+    if (aggregate_flow != NULL) {
+        aggregate_flow->end_mono_time_ts = bpf_ktime_get_ns();
+        aggregate_flow->flags |= flags;
+        if (aggregate_flow->flow_rtt < rtt) {
+            aggregate_flow->flow_rtt = rtt;
         }
```
jotak, line 81: So this means we keep always the highest srtt per flow ... which seems legit to me as we're generally interested in finding high latencies - but that might be something to keep in mind for documentation, if we document that level of details.

bpf/rtt_tracker.h
```diff
@@ -1,143 +1,161 @@
...
+    dport = bpf_ntohs(tcp.dest);
+    id->src_port = sport;
+    id->dst_port = dport;
+    set_flags(&tcp, flags);
+    id->transport_protocol = IPPROTO_TCP;
+}
+
+static inline int rtt_lookup_and_update_flow(flow_id *id, u16 flags, u64 rtt) {
+    flow_metrics *aggregate_flow = bpf_map_lookup_elem(&aggregated_flows, id);
+    if (aggregate_flow != NULL) {
```
jotak, line 76: You mentioned missed RTT when sampling is 1; could it be because the `aggregate_flow` might be null ? If that's the case, perhaps we can store temporarily the RTT in another map for that case where lookup returned null ?

bpf/rtt_tracker.h
```diff
@@ -1,143 +1,161 @@
...
-    u32 seq = bpf_ntohl(tcp->seq);
-    fill_flow_seq_id(seq_id, pkt, seq, false);
-    long ret = bpf_map_update_elem(&flow_sequences, seq_id, &pkt->current_ts, BPF_NOEXIST);
-    if (trace_messages && ret != 0) {
-        bpf_printk("err saving flow sequence record %d", ret);
-    }
-    return;
-}
+static inline int calculate_flow_rtt_tcp(struct sock *sk, struct sk_buff *skb) {
+	struct tcp_sock *ts;
```
jotak, line 92: nit: don't we have a linter for C code format? looks like there's a double tab here. Not important.

bpf/rtt_tracker.h
```diff
@@ -1,143 +1,161 @@
...
+        bpf_printk("error rtt track creating flow %d\n", ret);
     }
+
+    return 0;
 }
 
-static __always_inline void calculate_flow_rtt(pkt_info *pkt, u8 direction, void *data_end) {
-    flow_seq_id seq_id;
-    __builtin_memset(&seq_id, 0, sizeof(flow_seq_id));
+SEC("fentry/tcp_rcv_established")
```
jotak, line 145: for my own knowledge, what's "fentry/tcp_rcv_established" compared to "kprobe/tcp_rcv_established" ?

bpf/rtt_tracker.h
msherif1234: Yes I will and this is something I am debating with and waiting on rhel folks input other implementations aggregate rtt value I think getting the max is appropriate but I need confirmation and ofc we will reflect that in the doc and rtt blog

bpf/rtt_tracker.h
msherif1234: if the update path fails u will notice in the caller function it will create new flow with rtt value that could possible because of perCPU hash where the TC flow was recorded against one cpu but this hook fired on different cpu
I think with heavy sampling those hooks aren't getting chance to kick in at least that is my explanation atm

bpf/rtt_tracker.h
msherif1234: not really we don't have a C linter in the mix probably something nice to have I can look for one later, fixed this instance

bpf/rtt_tracker.h
msherif1234: `kprobe` inserted in the function while `fentry` stands for function entry everytime the function is called it will fire similarly there `fexit`the reason I am using double hook is following BCC implementation to make sure we never miss rtt update

msherif1234:
removed kprobe and just stayed with fentry only which more advanced kprobe and light weight

Amoghrd:
Works as expected with latest commits.

PR #278, opened 2024-02-26
Share errors counter across components

jotak:
- All errors can now be found from metric "errors_total"
- Use error name and exporter as labels

image


jotak:
/approve
thx

PR #281, opened 2024-02-27
NETOBSERV-1533: refine metrics for dashboard creation

jotak:
- Use just two shared metrics for eviction counters: on for eviction events and one for flows; shared across ringbuf/map implementations and labelled as such
- Report more errors via metrics

### DIFF
--- a/pkg/ebpf/tracer.go
+++ b/pkg/ebpf/tracer.go
@@ -595,9 +595,6 @@ func NewPacketFetcher(
 		return nil, fmt.Errorf("loading and assigning BPF objects: %w", err)
 	}
 
-	if err != nil {
-		return nil, err
-	}
 	// read packets from igress+egress perf array
 	packets, err := perf.NewReader(objects.PacketRecord, os.Getpagesize())
 	if err != nil {
--- a/pkg/ebpf/tracer.go
+++ b/pkg/ebpf/tracer.go
@@ -412,7 +412,7 @@ func (m *FlowFetcher) ReadRingBuf() (ringbuf.Record, error) {
 // TODO: detect whether BatchLookupAndDelete is supported (Kernel>=5.6) and use it selectively
 // Supported Lookup/Delete operations by kernel: https://github.com/iovisor/bcc/blob/master/docs/kernel-versions.md
 // Race conditions here causes that some flows are lost in high-load scenarios
-func (m *FlowFetcher) LookupAndDeleteMap(c *metrics.ErrorCounter) map[BpfFlowId][]BpfFlowMetrics {
+func (m *FlowFetcher) LookupAndDeleteMap(errs *metrics.ErrorCounter) map[BpfFlowId][]BpfFlowMetrics {
 	flowMap := m.objects.AggregatedFlows
 
 	iterator := flowMap.Iterate()
@@ -426,7 +426,7 @@ func (m *FlowFetcher) LookupAndDeleteMap(c *metrics.ErrorCounter) map[BpfFlowId]
 		if err := flowMap.Delete(id); err != nil {
 			log.WithError(err).WithField("flowId", id).
 				Warnf("couldn't delete flow entry")
-			c.ForError("CannotDeleteFlows").Inc()
+			errs.WithErrorName("flow-fetcher", "CannotDeleteFlows").Inc()
 		}
 		// We observed that eBFP PerCPU map might insert multiple times the same key in the map
 		// (probably due to race conditions) so we need to re-join metrics again at userspace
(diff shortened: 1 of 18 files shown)


jotak:
@msherif1234 as I'm working on the dashboard and I'm bringing a few changes here.
Here's a screenshot of what I get with these metrics:

Some text counters in the top bar:
Capture d’écran du 2024-02-27 16-11-07

Some ebpf stats:
Capture d’écran du 2024-02-27 16-11-13

(something looks wrong in the eviction rate when it evicts because of full map)

pkg/metrics/metrics.go
```diff
@@ -95,6 +89,11 @@ var (
 		"error",
 		"exporter",
 	)
+	// MapSize = defineMetric(
+	// 	"ebpf_map_size",
+	// 	"size of the eBPF maps",
+	// 	TypeGauge,
+	// )
```
jotak, line 96: I tried to create a gauge to track the hashmap size but I didn't find a function to get the used size in bpflib...

msherif1234: we can do dummy iterate like the one we use today and inside increment the map size ?

jotak: i'd prefer to avoid any additional processing .. no worries I'll delete this commented code

pkg/flow/tracer_map.go
```diff
@@ -44,17 +44,18 @@ func NewMapTracer(fetcher mapFetcher, evictionTimeout, staleEntriesEvictTimeout
...
 		errors:                     m.GetErrorsCounter(),
 	}
 }
 
 // Flush forces reading (and removing) all the flows from the source eBPF map
 // and sending the entries to the next stage in the pipeline
-func (m *MapTracer) Flush() {
+func (m *MapTracer) Flush(reason string) {
 	m.evictionCond.Broadcast()
+	m.evictionCounter.ForSourceAndReason("hashmap", reason).Inc()
```
jotak, line 58: @msherif1234 I think what I did here doesn't work: the eviction counter is incremented everytime `Flush` is called, but it seems to be called many times before sync.Condition makes it go to the actual evict. I'm not sure why and I wonder if this is normal. But to "fix" this counter I would need to move the increment back where you set it before, in `evictFlows`, but the downside is that we loose then the `reason` label that I wanted to have

jotak: Look at this: there's a peak to 300K evictions / second due to this increase, while the flows per second is only at 3.5K. Isn't it fishy?

Capture d’écran du 2024-02-27 16-11-25

pkg/metrics/metrics.go
msherif1234: me2 :)

jotak:
@msherif1234 I think this task is done I've got all metrics needed to build a nice dashboard - can you take a second look & review please?

pkg/ebpf/tracer.go
```diff
@@ -411,27 +412,30 @@ func (m *FlowFetcher) ReadRingBuf() (ringbuf.Record, error) {
...
+			log.WithError(err).WithField("flowId", id).Warnf("couldn't delete flow entry")
+			met.Errors.WithErrorName("flow-fetcher", "CannotDeleteFlows").Inc()
 		}
 		// We observed that eBFP PerCPU map might insert multiple times the same key in the map
 		// (probably due to race conditions) so we need to re-join metrics again at userspace
 		// TODO: instrument how many times the keys are is repeated in the same eviction
 		flows[id] = append(flows[id], metrics...)
 	}
+	met.BufferSizeGauge.WithBufferName("hashmap-total").Set(float64(count))
+	met.BufferSizeGauge.WithBufferName("hashmap-unique").Set(float64(len(flows)))
```
msherif1234, line 438: aren't those counters are dup to what we we already have in tracer_map ?

pkg/metrics/metrics.go
```diff
@@ -46,54 +46,62 @@ func defineMetric(name, help string, t metricType, labels ...string) MetricDefin
...
+		"Exported batches",
 		TypeCounter,
 		"exporter",
 	)
-	samplingRateSeconds = defineMetric(
-		"sampling_rate_seconds",
-		"Sampling rate seconds",
+	samplingRate = defineMetric(
+		"sampling_rate",
+		"Sampling rate",
```
msherif1234, line 96: adding the unis was in the metrics guide line u shared with me ?

msherif1234:
overall looks great to me small comments/questions inline up to you to address nothing critical
/lgtm

pkg/metrics/metrics.go
jotak: but this one is the sampling rate (ratio) like 1:50 or 1:1 it's not measuring time

pkg/ebpf/tracer.go
jotak: it's different , the counters are used to compute rates, like number of flows evicted per second, but this allows to get the size of the maps. In fact (to answer your next question) they have been incredibly useful to me as it's thanks to these gauges that I understood exactly what was going wrong [here](https://github.com/netobserv/netobserv-ebpf-agent/pull/283) with LookupAndDelete: the "hashmap-total" was growing to 100K elements ie. the full map size even when it was not supposed to be full, while the "hashmap-unique" gauge was much smaller, something like 2K. So that's what led me to find the issue about deleting within the iteration.

PR #286, opened 2024-03-04
NETOBSERV-1637: OVS monitoring ebpf hook

msherif1234:
## Description

OVS monitoring eBPF hook feature

ebpf configs to enable ovs tracking from operator

```yaml
      advanced:
        env:
          ENABLE_OVS_MONITORING: "true"
```

```sh
- bpftool perf show -p
[{
        "pid": 2854,
        "fd": 8,
        "prog_id": 143,
        "fd_type": "kprobe",
        "func": "psample_sample_packet",
        "offset": 0
    }
]
- bpftool map dump id 258
"key": {
            "eth_protocol": 2048,
            "direction": 0,
            "src_mac": [2,220,231,139,148,213
            ],
            "dst_mac": [10,88,10,128,2,12
            ],
            "src_ip": [0,0,0,0,0,0,0,0,0,0,255,255,10,128,2,2
            ],
            "dst_ip": [0,0,0,0,0,0,0,0,0,0,255,255,10,128,2,12
            ],
            "src_port": 56546,
            "dst_port": 8080,
            "transport_protocol": 6,
            "icmp_type": 0,
            "icmp_code": 0,
            "if_index": 2
        },
        "values": [{
                "cpu": 0,
                "value": {
                    "packets": 1,
                    "bytes": 74,
                    "start_mono_time_ts": 6218828496667,
                    "end_mono_time_ts": 6218828496667,
                    "flags": 2,
                    "errno": 0,
                    "dscp": 0,
                    "pkt_drops": {
                        "packets": 0,
                        "bytes": 0,
                        "latest_flags": 0,
                        "latest_state": 0,
                        "latest_drop_cause": 0
                    },
                    "dns_record": {
                        "id": 0,
                        "flags": 0,
                        "latency": 0,
                        "errno": 0
                    },
                    "flow_rtt": 0,
                     ovs_dp_keys": [[0,0,0,0,12,0,255,238
                        ],[0,0,0,0,0,0,0,0
                        ],[0,0,0,0,0,0,0,0
                        ],[0,0,0,0,0,0,0,0
                        ]
```

## Dependencies
- new kernel update with OVS hook
- new version of OVS
- new version of OVN
- https://github.com/ovn-org/ovn-kubernetes/pull/4596
- https://github.com/netobserv/network-observability-operator/pull/664
- https://github.com/netobserv/flowlogs-pipeline/pull/669
- https://github.com/netobserv/network-observability-console-plugin/pull/535

## Checklist


go.mod
```diff
@@ -121,16 +125,18 @@ require (
...
 	lukechampine.com/uint128 v1.2.0 // indirect
-	sigs.k8s.io/controller-runtime v0.15.1 // indirect
+	sigs.k8s.io/controller-runtime v0.17.0 // indirect
 	sigs.k8s.io/json v0.0.0-20221116044647-bc3834ca7abd // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
-	sigs.k8s.io/yaml v1.3.0 // indirect
+	sigs.k8s.io/yaml v1.4.0 // indirect
 )
+
+replace github.com/ovn-org/ovn-kubernetes/go-controller => github.com/npinaeva/ovn-kubernetes/go-controller v0.0.0-20240514113629-6445cb329c05
```
msherif1234, line 142: FIXME: need to be replaced by the right ovnk version once the changes are merged to master

pkg/decode/decode_protobuf.go
```diff
@@ -136,6 +137,15 @@ func RecordToMap(fr *flow.Record) config.GenericMap {
...
 		out["TimeFlowRttNs"] = fr.TimeFlowRtt.Nanoseconds()
 	}
+
+	var ovsMetadatas []string
+	for _, md := range fr.OvsMonitorMD {
+		if md != "" {
+			ovsMetadatas = append(ovsMetadatas, md)
+		}
+	}
+	out["OvsMonitorMDs"] = ovsMetadatas
```
jpinsonneau, line 147: @msherif1234 as discussed I would like to merge metadatas in a single field as we did for Interfaces and Directions

msherif1234: done

bpf/rtt_tracker.h
```diff
@@ -89,12 +27,12 @@ static inline int rtt_lookup_and_update_flow(flow_id *id, u16 flags, u64 rtt) {
 }
 
 static inline int calculate_flow_rtt_tcp(struct sock *sk, struct sk_buff *skb) {
+    u8 dscp = 0, protocol = 0;
     struct tcp_sock *ts;
     u16 family, flags = 0;
```
jotak, line 32: could you initialize family to 0?
It's later passed down into `core_fill_in_l2` which does the initialization but not in all paths

msherif1234: its initialized at line 32

bpf/utils.h
```diff
@@ -337,4 +338,129 @@ static inline bool check_and_do_flow_filtering(flow_id *id, u16 flags) {
...
+
+    bpf_probe_read(&eth, sizeof(eth), (struct ethhdr *)(skb_head + skb_mac_header));
+    __builtin_memcpy(id->dst_mac, eth.h_dest, ETH_ALEN);
+    __builtin_memcpy(id->src_mac, eth.h_source, ETH_ALEN);
+    id->eth_protocol = bpf_ntohs(eth.h_proto);
+    if (id->eth_protocol == ETH_P_IP) {
+        *family = AF_INET;
+    } else if (id->eth_protocol == ETH_P_IPV6) {
+        *family = AF_INET6;
+    }
```
jotak, line 357: previously, `family` was obtained that way: `family = BPF_CORE_READ(sk, __sk_common.skc_family);` in code for RTT ... I wonder if this change could impact the bug #382 , where in some situations it seemed like ipv4 aren't correctly read

msherif1234: I will check on the issue with RTT once this is merged see if I can repo but its possible if skb family wasn't set it will default to v4 we will see

bpf/utils.h
```diff
@@ -337,4 +338,129 @@ static inline bool check_and_do_flow_filtering(flow_id *id, u16 flags) {
...
+
+static inline void core_fill_in_tcp(struct sk_buff *skb, flow_id *id, u16 *flags) {
+    u16 skb_transport_header = BPF_CORE_READ(skb, transport_header);
+    u8 *skb_head = BPF_CORE_READ(skb, head);
+    struct tcphdr tcp;
+    u16 sport, dport;
+
+    __builtin_memset(&tcp, 0, sizeof(tcp));
+
+    bpf_probe_read(&tcp, sizeof(tcp), (struct tcphdr *)(skb_head + skb_transport_header));
```
jotak, line 401: I just noticed that in utils.h [here](https://github.com/netobserv/netobserv-ebpf-agent/blob/main/bpf/utils.h#L47-L84) , we do a similar extraction of the headers but always with a protective block to ensure we're not trying to read more than `skb->data_end`. Here and below we don't do that, but should we?

msherif1234: I don't think data_end is populated for the rtt and psample hook CORE function should be safe otherwise the verifier would screamed at us

pkg/agent/config.go
```diff
@@ -221,6 +220,10 @@ type Config struct {
 	// FilterTCPFlags is the TCP flags to filter flows.
 	// possible values are: SYN, SYN-ACK, ACK, FIN, RST, PSH, URG, ECE, CWR, FIN-ACK, RST-ACK
 	FilterTCPFlags string `env:"FILTER_TCP_FLAGS"`
+	// EnableNetworkEventsMonitoring enables monitoring network plugin events, default is false.
+	EnableNetworkEventsMonitoring bool `env:"ENABLE_NETWORK_EVENTS_MONITORING" envDefault:"false"`
+	// NetworkEventsMonitoringGroupID to allow ebpf hook to process samples for specific groupID and ignore the rest
+	NetworkEventsMonitoringGroupID int `env:"NETWORK_EVENTS_MONITORING_GROUP_ID" envDefault:"10"`
```
jotak, line 226: what's exactly this group ID, how are we supposed to know what to set?

bpf/network_events_monitoring.h
```diff
@@ -0,0 +1,151 @@
...
+    // filter out none matching samples with different groupid
+    int group_id = BPF_CORE_READ(group, group_num);
+    if (group_id != network_events_monitoring_groupid) {
+        return 0;
+    }
+
+    return trace_network_events(skb, md);
+}
+/*
+SEC("kprobe/psample_sample_packet")
```
jotak, line 133: is it planned to uncomment later?

bpf/rtt_tracker.h
jotak: flags yes, but family no: `u16 family, flags = 0;`

bpf/network_events_monitoring.h
msherif1234: its planned for the new kernel whick is `REHL 9.6` with the planned kernel for 9.4 we have `rh_` version of the hook because they can't break kernel apis but in 9.6 `rh_` will be gone that why I ahve this with kernel version check for the future

pkg/pbflow/proto.go
```diff
@@ -94,6 +97,18 @@ func FlowToPB(fr *flow.Record) *Record {
 		pbflowRecord.Network.SrcAddr = &IP{IpFamily: &IP_Ipv4{Ipv4: flow.IntEncodeV4(fr.Id.SrcIp)}}
 		pbflowRecord.Network.DstAddr = &IP{IpFamily: &IP_Ipv4{Ipv4: flow.IntEncodeV4(fr.Id.DstIp)}}
 	}
+	if s != nil {
+		for _, metadata := range fr.Metrics.NetworkEvents {
+			if !flow.AllZerosMetaData(metadata) {
+				if md, err := s.DecodeCookie8Bytes(metadata); err == nil {
```
jotak, line 103: I'm not convinced that doing this decoding within the protobuf encoding is the best option we have, it doesn't have to be tied to the transport data model - also I'm not sure if the DirectFLP mode uses pb, I don't think? So that would mean to duplicate in several places the decoding part.

I don't think it's a blocker for the current work, but maybe to consider as a follow-up? For instance, we could create a new stage in `buildAndStartPipeline` (in agent.go) to do the decoding.

msherif1234: I was looking at the direct-flp path and it does populate the cm directly but yeah we can follow up with further enhancement as needed that was the best I can come up with with less risk

pkg/agent/config.go
jotak: so this is somehow tied to an OVN configuration that is done somewhere else?

msherif1234: yeah in ovn and currently hardcoded to `10`

msherif1234: here is the slack thread about it for reference
```sh
amorenoz
@Nadia Pinaeva @mmahmoud
 I'd keep it simple for the time being, how about this:
ovnk-observ-library:
The API allows the user to provide a group and an external_id, e.g: configureCollector(id, 10, "ebpfAgent")
The API allows the user to clean up the configuration when it is closed normally.
The library checks if the entry already exists:
If it exists, it may mean that ebpfAgent crashed and did not clean up resources, the library checks if the external id is the same.
If it is, then the configuration is kept and returns OK (or maybe a code indicating that it already existed but it's not an error)
If it's not, return an error, someone stole the group from NetObserv, it's a bug [1]
If not, the library configures and returns OK.
ebpfAgent
Hardcodes and documents the group id that it'll use
Configures it via library
ebpf code just checks against that number.
[1] We could be smarter and nave ebpfAgent generate another number and retry here but that'd mean  ebpfAgent has to deal with dynamic ids which is a bit more complicated.
Apart from error cases like [1], in the future we may want  to configure more than one collector, i.e: more than one group_id, for NetObserv. So at some point we may need to support injecting them dynamically into the ebpf agent through a MAP or something. But I think right now is not strictly needed and we could save some devel time here.
```

jotak (approved):
LGTM thanks @msherif1234 !

PR #291, opened 2024-03-06
NETOBSERV-1471 gRPC export for packet capture

jpinsonneau:
## Description

Add gRPC export capability for packet capture.
Since I moved flow grpc to `flowgrpc` package, we will need to update FLP at the same time

I also took the opportunity to reuse the target host / port from flow config since it doesn't make sense to have both here. 

## Dependencies

n/a

## Checklist

  * [X] No QE (e.g. trivial change with high reviewer's confidence, or per agreement with the QE team).


pkg/agent/config.go
```diff
@@ -163,8 +164,6 @@ type Config struct {
 	// PCAFilters set the filters to determine packets to filter using Packet Capture Agent (PCA). It is a comma separated set.
 	// The format is [protocol], [port number] Example: PCA_FILTER = "tcp,80". Currently, we support 'tcp','udp','sctp' for protocol.
 	PCAFilters string `env:"PCA_FILTER"`
-	// PCAServerPort is the port PCA Server starts at, when ENABLE_PCA variable is set to true.
-	PCAServerPort int `env:"PCA_SERVER_PORT" envDefault:"9990"`
```
jotak, line 167: I remember we had this discussion with Shachee previously, that reusing the `FLOWS_TARGET_PORT` env for pcap was misleading, and for two reasons: first because it has "FLOWS" in the name, and second because in the case of the TCP impl, if I remember correctly, this isn't a target port (ie. a remote host port) but it's a server port, uses to listen for clients who want to poll data. So that's why we ended up with this new env.

pkg/exporter/grpc_packets.go
```diff
@@ -0,0 +1,78 @@
...
+
+type GRPCPacketProto struct {
+	hostIP     string
+	hostPort   int
+	clientConn *grpc.ClientConnection
+}
+
+var gplog = logrus.WithField("component", "packet/GRPCPackets")
+
+// WritePacket writes the given packet data out to the file.
```
msherif1234, line 24: this doesn't write to a file ?

pkg/exporter/grpc_packets.go
```diff
@@ -0,0 +1,78 @@
...
+				packetTimestamp := packet.Time
+				if len(packetStream) != 0 {
+					captureInfo := gopacket.CaptureInfo{
+						Timestamp:     packetTimestamp,
+						CaptureLength: len(packetStream),
+						Length:        len(packetStream),
+					}
+					err := writeGRPCPacket(captureInfo, packetStream, p.clientConn)
+					if err != nil {
+						gplog.Fatal(err)
```
msherif1234, line 72: do u really need to assert when this error happens or just log an error and try another packet ?

pkg/exporter/tcp_packets.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func (p *PCAPStream) ExportTCPPackets(in <-chan []*flow.PacketRecord) {
+	//Create handler by opening PCAP stream - Write 24 byte size PCAP File Header
+	_, err := p.clientConn.Write(GetPCAPFileHeader(snapshotlen, layers.LinkTypeEthernet))
+	if err != nil {
+		tplog.Fatal(err)
+	}
+	for packetRecord := range in {
+		if len(packetRecord) > 0 {
```
msherif1234, line 80: same here I don't think we need len check

pkg/exporter/grpc_packets.go
```diff
@@ -0,0 +1,78 @@
...
+			for _, packet := range packetRecord {
+				packetStream := packet.Stream
+				packetTimestamp := packet.Time
+				if len(packetStream) != 0 {
+					captureInfo := gopacket.CaptureInfo{
+						Timestamp:     packetTimestamp,
+						CaptureLength: len(packetStream),
+						Length:        len(packetStream),
+					}
+					err := writeGRPCPacket(captureInfo, packetStream, p.clientConn)
```
msherif1234, line 70: I wonder if we can define grpc as array of pcap and write all the record at once instead of writing every packet this will probably be more efficient WDYT?

proto/packet.proto
```diff
@@ -0,0 +1,19 @@
...
+
+import "google/protobuf/any.proto";
+
+service Collector {
+  rpc Send (Packet) returns (CollectorReply) {}
+}
+
+// The request message containing the Packet
+message Packet {
+  google.protobuf.Any pcap = 1;
```
msherif1234, line 15: if we decided to send multiple pcaps then we need to add `repeated` at the definition above

proto/packet.proto
```diff
@@ -0,0 +1,19 @@
...
+}
+
+// intentionally empty
+message CollectorReply {}
```
msherif1234, line 19: nit: pls add new line here

examples/packetcapture-dump/README.md
```diff
@@ -14,7 +14,7 @@ go build -mod vendor -o bin/packetcapture-client examples/packetcapture-dump/cli
 ```
 Start the agent using:
 ```bash
-sudo PCA_SERVER_PORT=9990 ENABLE_PCA=true PCA_FILTER=tcp,22 ./bin/netobserv-ebpf-agent
+sudo EXPORT=tcp FLOWS_TARGET_PORT=9990 ENABLE_PCA=true PCA_FILTER=tcp,22 ./bin/netobserv-ebpf-agent
```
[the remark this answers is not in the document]
jpinsonneau: The example was on tcp so I wanted to keep it as is. I can add another one for gRPC if you want :+1:

pkg/agent/config.go
```diff
@@ -34,8 +34,9 @@ type Config struct {
 	// in the AgentID field of each flow. Accepted values are: any (default), ipv4, ipv6.
 	// If the AgentIP configuration property is set, this property has no effect.
 	AgentIPType string `env:"AGENT_IP_TYPE" envDefault:"any"`
-	// Export selects the flows' exporter protocol. Accepted values are: grpc (default), kafka,
-	// ipfix+udp, ipfix+tcp or direct-flp.
+	// Export selects the exporter protocol.
+	// Accepted values for Flows are: grpc (default), kafka, ipfix+udp, ipfix+tcp or direct-flp.
+	// Accepted values for Packets are: grpc (default) or tcp
```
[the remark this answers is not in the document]
jpinsonneau: Previously, PCA was only supporting tcp and required its own configs prefixed with PCA_
Now [I'm trying to merge with flow configs](https://github.com/netobserv/netobserv-ebpf-agent/pull/291#discussion_r1521064391) and support both gRPC + tcp

Don't you think it's worth ?

pkg/exporter/grpc_packets.go
```diff
@@ -0,0 +1,78 @@
...
+	return &GRPCPacketProto{
+		hostIP:     hostIP,
+		hostPort:   hostPort,
+		clientConn: clientConn,
+	}, nil
+}
+
+func (p *GRPCPacketProto) ExportGRPCPackets(in <-chan []*flow.PacketRecord) {
+	for packetRecord := range in {
+		if len(packetRecord) > 0 {
```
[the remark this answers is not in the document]
jpinsonneau: That was the way it's done in original tcp packet capture; that's said, if the range ensure the array is not empty, I can remove this

pkg/exporter/grpc_packets.go
jpinsonneau: I can give a try indeed. What would be the maximum acceptable total size here ?

A single packet can be huge so I may need to check about the next(s) before sending

pkg/agent/config.go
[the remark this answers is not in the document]
msherif1234: why we maintain two exporter if we can just use grpc unless there is some known limitation that tcp fix but if we have no answer I think we can just use gRPC

pkg/exporter/grpc_packets.go
jpinsonneau: For the collector side, it doesn't make sense to get half of a packet here. It would require extra work to sync and avoid concurrency

We should keep this simple and write packets entirely; but allowing array of them with a reasonnable upper limit

pkg/agent/config.go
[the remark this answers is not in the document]
jpinsonneau: @jotak @shach33 thoughts ?

pkg/exporter/grpc_packets.go
msherif1234: are we ok to limit the packet header size from the ebpf to say 256 bytes max if the packet is fragmented or has encrypted data it has no value copy all of that to user space WDYT ?

jpinsonneau: As we discussed today, it's still interesting to capture the payload for non encrypted packets. Giving the ability to get only the firsts 256 bytes is still interesting if wireshark manages that seamlessly but in that case, flows already cover these informations.

vendor/github.com/netobserv/flowlogs-pipeline/pkg/pipeline/ingest/ingest_grpc.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/netobserv/flowlogs-pipeline/pkg/operational"
 	"github.com/netobserv/flowlogs-pipeline/pkg/pipeline/utils"
 	"github.com/netobserv/netobserv-ebpf-agent/pkg/decode"
-	"github.com/netobserv/netobserv-ebpf-agent/pkg/grpc"
+	grpc "github.com/netobserv/netobserv-ebpf-agent/pkg/grpc/flow"
```
jpinsonneau, line 12: :point_up_2: that cross import will require FLP update

pkg/agent/config.go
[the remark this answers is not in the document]
shach33: I feel as @jpinsonneau  mentioned, initially with PCA the support was intended only for tcp. Merging this would make using the tool more structured.  
@msherif1234 The reason to have a it separate from grpc was to export in pcap.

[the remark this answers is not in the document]
msherif1234: with grpc we are still producing pcaps from the CLI repo too right ? my worry is we have a mode that is not used but I am not strongly against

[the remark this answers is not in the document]
shach33: If we still continue to export packets as pcap via grpc, i feel we should move forward with the change.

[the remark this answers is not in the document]
msherif1234: @jpinsonneau pls confirm ?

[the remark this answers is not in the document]
jpinsonneau: Indeed the same bytes are sent either with TCP or gRPC
both produces pcap files in the CLI

pkg/agent/config.go
jpinsonneau: FYI I addressed this in that way in:
https://github.com/netobserv/netobserv-ebpf-agent/pull/291/commits/8cbecafbadef42631196dd60e49b7e5547464859
https://github.com/netobserv/netobserv-ebpf-agent/pull/291/commits/3e86b43ccc3c9fbb2d89804537b5110b002a8ba0

jotak: idk it still bother me to mix server and client config into a single configuration ; you kept "target" which IMO is still misleading for a server configuration, it really looks like we are configuring a client. idk if it's just me, @msherif1234 what do you think? I defer to you :-)

jotak: well if we get rid of pcap w/ tcp , hence we have no server mode anyway, then I guess it answers the question

jpinsonneau:
Rebased + removed tcp packet capture
https://github.com/netobserv/netobserv-ebpf-agent/pull/291/commits/6e65e79a0050842f22a82466d71bba97e02550ed

pkg/agent/config.go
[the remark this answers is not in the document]
jpinsonneau: I removed tcp and updated example
https://github.com/netobserv/netobserv-ebpf-agent/pull/291/commits/6e65e79a0050842f22a82466d71bba97e02550ed

pkg/agent/config.go
jpinsonneau: As mention above I changed the config to get rid of these mentions. That's more generic for future implementations.
On top of that I removed tcp export for packet capture.

Let me know if the current state match your expectations

proto/packet.proto
jpinsonneau: Can we keep that for a followup improvment ? That raise more questions about the total size of the message.

The goal here was just to implement gRPC for security reasons to productize CLI.

pkg/agent/config.go
```diff
@@ -174,4 +177,30 @@ type Config struct {
...
+	/* Deprecated configs are listed below this line
+	 * See manageDeprecatedConfigs function for details
+	 */
+
+	// Deprecated FlowsTargetHost replaced by TargetHost
+	FlowsTargetHost string `env:"FLOWS_TARGET_HOST"`
+	// Deprecated FlowsTargetPort replaced by TargetPort
+	FlowsTargetPort int `env:"FLOWS_TARGET_PORT"`
+	// Deprecated PCAServerPort replaced by TargetPort
+	PCAServerPort int `env:"PCA_SERVER_PORT"`
```
msherif1234, line 190: if we have case like we have both agent and pca running at the same node reaching out to host but using different port with the above proposed config that seems impossible right ?
i.e with this new config changes we can run either agent or pca but not both at the same time as it was b4 this change ?

pkg/agent/config.go
```diff
@@ -174,4 +177,30 @@ type Config struct {
...
+}
+
+func manageDeprecatedConfigs(cfg *Config) {
+	if len(cfg.FlowsTargetHost) > 0 {
```
msherif1234, line 194: nit: u are doing `!=0` below so why not be consistent ?

pkg/exporter/grpc_packets.go
```diff
@@ -0,0 +1,79 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/google/gopacket"
+	"github.com/sirupsen/logrus"
+	"google.golang.org/protobuf/types/known/anypb"
+
+	"github.com/netobserv/netobserv-ebpf-agent/pkg/flow"
+	grpc "github.com/netobserv/netobserv-ebpf-agent/pkg/grpc/packet"
```
msherif1234, line 12: nit: this is the import order used in the agent
1- standard libs
2- netobserv pkg
3- external packages
this applies here and in all newly created files

pkg/exporter/grpc_packets.go
```diff
@@ -0,0 +1,79 @@
...
+			packetTimestamp := packet.Time
+			if len(packetStream) != 0 {
+				captureInfo := gopacket.CaptureInfo{
+					Timestamp:     packetTimestamp,
+					CaptureLength: len(packetStream),
+					Length:        len(packetStream),
+				}
+				err := writeGRPCPacket(captureInfo, packetStream, p.clientConn)
+				if err != nil {
+					gplog.Error(err)
```
msherif1234, line 71: lets aggregate the errors here and return error in this function and use the return err from here in the caller not the err from `StartGRPCPacketSend`

pkg/agent/config.go
jpinsonneau: You can't have both `flow` + `pca` capture as it rely on `ENABLE_PCA` bool config
https://github.com/netobserv/netobserv-ebpf-agent/blob/main/cmd/netobserv-ebpf-agent.go#L56

pkg/agent/config.go
jpinsonneau: Host is a `string` whereas Port is a `number`

pkg/exporter/grpc_packets.go
jpinsonneau: Sure, https://github.com/netobserv/netobserv-ebpf-agent/pull/291/commits/111c915245d415d29c9cd9710ad79e323207b51e

pkg/agent/config.go
msherif1234: i meant len() != 0 as oppose to len() > 0

pkg/exporter/grpc_packets.go
msherif1234: if u look at the caller it return error from startGRPC right but not reflecting errors because of export 
so what I am suggesting return an error here and inside the loop just log and aggr the errors

pkg/grpc/packet/grpc_test.go
```diff
@@ -0,0 +1,87 @@
+package pktgrpc
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/mariomac/guara/pkg/test"
+	"github.com/netobserv/netobserv-ebpf-agent/pkg/pbpacket"
```
msherif1234, line 9: nit: import order

pkg/grpc/packet/server.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	"context"
+	"fmt"
+	"net"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/reflection"
+
+	"github.com/netobserv/netobserv-ebpf-agent/pkg/pbpacket"
```
msherif1234, line 11: here too

examples/packetcapture-dump/client/packetcapture-client.go
```diff
@@ -20,17 +20,23 @@ package main
 import (
 	"flag"
 	"fmt"
-	"net"
 	"os"
 	"time"
+
+	"github.com/google/gopacket/layers"
```
msherif1234, line 26: import order

msherif1234:
I am good with the changes few nit left other than that LGTM

pkg/exporter/grpc_packets.go
jpinsonneau: is that what you had in mind ? 

https://github.com/netobserv/netobserv-ebpf-agent/pull/291/commits/0ccc69babe640244e8b0234d9f955e3ffd9a2bbc

pkg/grpc/packet/grpc_test.go
jpinsonneau: done in https://github.com/netobserv/netobserv-ebpf-agent/pull/291/commits/beef4dd954e19bba5aec090685e889b2ee2e233f

pkg/grpc/packet/server.go
jpinsonneau: done in https://github.com/netobserv/netobserv-ebpf-agent/pull/291/commits/beef4dd954e19bba5aec090685e889b2ee2e233f

examples/packetcapture-dump/client/packetcapture-client.go
jpinsonneau: done in https://github.com/netobserv/netobserv-ebpf-agent/pull/291/commits/beef4dd954e19bba5aec090685e889b2ee2e233f

msherif1234:
pls remember to update the operator
https://github.com/netobserv/network-observability-operator/blob/1cd8e7670e47ee248965c72e5fc9bc8709fd615f/controllers/ebpf/agent_controller.go#L35

pkg/agent/config.go
jotak: Well, sorry but, now that it's clearly only a client connection there's no reason to remove "target" from the names? It was inconsistent if we were mixing client & server, but now that we don't, it's fine and more precise than just "host" / "port"
Anyway I don't want to block this PR indefinitely, we can do that later

jotak (approved):

jpinsonneau:
> pls remember to update the operator https://github.com/netobserv/network-observability-operator/blob/1cd8e7670e47ee248965c72e5fc9bc8709fd615f/controllers/ebpf/agent_controller.go#L35

Here is the PR:
https://github.com/netobserv/network-observability-operator/pull/600

PR #290, opened 2024-03-06
Bump FLP for direct-flp

jotak:
fix for latest breaking changes in FLP

PR #295, opened 2024-03-12
NETOBSERV-1514: drop go1.20 support

jotak:


msherif1234:
pls update here too https://github.com/netobserv/netobserv-ebpf-agent/blob/main/scripts/generators.Dockerfile#L3

jotak:
> pls update here too https://github.com/netobserv/netobserv-ebpf-agent/blob/main/scripts/generators.Dockerfile#L3

why? it's dropping support for 1.20 and starts testing on 1.22 but it doesn't change the current used version in builds which is 1.21

msherif1234:
> > pls update here too https://github.com/netobserv/netobserv-ebpf-agent/blob/main/scripts/generators.Dockerfile#L3
> 
> why? it's dropping support for 1.20 and starts testing on 1.22 but it doesn't change the current used version in builds which is 1.21

currently it uses `1.21.3` while u going to `1.21.7` so just to keep all in sync ?

jotak:
hmm I see .. it's annoying to have this "toolchain 1.21.7" mention in go.mod, I would say any 1.21 version should be supported for the dev team we shouldn't have to stick to a particular z-stream on our local machines

msherif1234:
/lgtm
e2e errors seems to be baseline issue after recent FLP update #290

PR #297, opened 2024-03-14
Fix issue with inconsistent GenericMap output

jotak:
## Description

The FLP GenericMap provided by the agent differs in types depending on the exporter being used, especially with DirectFLP each records size is smaller than when used with protobuf, due to protobuf using uint32 almost everywhere. This is not a problem in most cases as FLP generally converts that in JSON, but it breaks the FLP IPFIX export which expects strict types.

This commit brings more consistency between the FLP-direct mode and the protobuf mode, by sharing the same GenericMap encoding function. Which means the protobuf decode is now done in two steps: first proto-to-flow.Record, then flow.Record-to-GenericMap.

## Dependencies

FLP update of the IPFIX exporter: https://github.com/netobserv/flowlogs-pipeline/pull/634

## Checklist


pkg/pbflow/proto.go
```diff
@@ -0,0 +1,202 @@
...
+		}
+		records = append(records, &Records{Entries: entries[:end]})
+		entries = entries[end:]
+	}
+	return records
+}
+
+// FlowToPB is an auxiliary function to convert a single flow record, as returned by the eBPF agent,
+// into a protobuf-encoded message ready to be sent to the collector via kafka
+func FlowToPB(fr *flow.Record) *Record {
```
msherif1234, line 34: curious why did u move this file from export to here ?

pkg/pbflow/proto.go
```diff
@@ -0,0 +1,202 @@
...
+	// IPv6 address
+	return &IP{IpFamily: &IP_Ipv6{Ipv6: nip}}
+}
+
+func pbIPToNetIP(ip *IP) net.IP {
+	if ip.GetIpv6() != nil {
+		return net.IP(ip.GetIpv6())
+	}
+	n := ip.GetIpv4()
+	return net.IPv4(
```
msherif1234, line 182: is it any cleaner like this
```go
byteArray := make([]byte, 4)
binary.NativeEndian.PutUint32(byteArray, n)
```
or is it possible to maye it byte type in pb like ipv6 so we can avoid all of this ?

msherif1234:
any cpu impact for the additional processing with ur scale test ?

msherif1234:
why this is marked as breaking change ?

pkg/pbflow/proto.go
jotak: it's also called from `decode` package now, and leaving it in `export` would create a cycle. The other option would be to move it to `decode`, I can do that if you prefer.

pkg/pbflow/proto.go
jotak: I can do your first suggestion (I'm not sure about the other, if the way it is was initially done for good reasons or not..)

But I guess I should either use LittleEndian or BigEndian but not Native, as that would make it arch-dependent (and thus fail on BE archs)

jotak:
> why this is marked as breaking change ?

That's a breaking change because the types in the GenericMap output change, as you can see here: https://github.com/netobserv/netobserv-ebpf-agent/pull/297/files#diff-5b98b198dc9b743f2c1c2d470d443d9ec2ed38b4ee15777cb857584ddacce657
For instance "Proto" was u8 in the inner model, cast as uint32 in protobuf, previously with the grpc exporter it was still uint32 in the GenericMap but now it's back to uint8 in that Map.

So, consumers who would cast explicitly to uint32 would now have to cast to uint8

jotak:
> any cpu impact for the additional processing with ur scale test ?

Slight increase of CPU in one of the tests (0.167 -> 0.172) and memory as well (671MB -> 679 MB) (i picked to one run that seems more relevant as it showed a similar number of flows per seconds). I'd say it's in reasonable bounds...

jotak:
/approve
marking no-qe as there's no visible output

PR #307, opened 2024-03-29
NETOBSERV-1478: Add eBPF Agent filtering capability

### LINKED ISSUE #266: Add feature to exclude ip from flow
As a user of this agent, it can overwhelm the flow pipeline (for k8s enriching) on very large clusters. It would be optimal, if I could add ip filtering in the netobserv-ebpf-agent so that I can pass only interesting flow to downstream flow pipline.

As a very hacky example https://github.com/netobserv/netobserv-ebpf-agent/commit/c55b0acbb0238b2f21beba022381ea87b9014f31

msherif1234:
## Description

This PR allow provisioning global filter to limit what goes to eBPF flows table
this also fix #266 

## Dependencies

n/a

## Checklist

* [X] Does this PR require product documentation?
  * [X] If so, make sure the JIRA epic is labelled with "documentation" and provides a description relevant for doc writers, such as use cases or scenarios. Any required step to activate or configure the feature should be documented there, such as new CRD knobs.
* [X] Is there anything else the QE team should know before testing? E.g: configuration changes, environment setup, etc.


bpf/flows_filter.h
```diff
@@ -0,0 +1,189 @@
...
+/*
+ * check if the flow match filter rule and return >= 1 if the flow is to be dropped
+ */
+static __always_inline int is_drop_flow(flow_id *id) {
+    struct filter_key_t key;
+    u8 len = 0, offset = 0;
+    int result = 0;
+
+    __builtin_memset(&key, 0, sizeof(key));
+    // for Ingress side we do srcIP CIDR matching and for Egress side we do dstIP CIDR matching
```
jotak, line 47: I'm trying to see if this allows us to work with these use cases:

- I want to get all outgoing traffic from my pod `foo`
- I want to get all ingoing traffic to my pod `foo`
- I want to get all traffic (in+out) of my pod `foo`

The assumption to check srcIP for Ingress and dstIP for Egress doesn't seem to work here? E.g. let's say my pod is 1.2.3.4 and the peer is 8.8.8.8

Flow A = (src=1.2.3.4, dst=8.8.8.8, dir=egress)
Flow B = (src=8.8.8.8, dst=1.2.3.4, dir=ingress)

I'll miss these two flows if I'm filtering for CIDR (or IP) 1.2.3.4/32, no ? Because in both cases we would try to match 8.8.8.8 instead of 1.2.3.4
I know with the different interfaces involved & dups, probably I would still get something, "thanks" to the duplicates... But I'm wondering if we shouldn't instead match both SRC and DST, regardless of the direction, when doing the IP matching: that way we get all the related flows, for all involved interfaces

jotak: huh, I just realize the rule is reversed for the IP matching. Ie for CIDR matching you check src for ingress and  dst for egress, but for IP matching you check **dst** for ingress and **src** for egress. So it's like CIDR is used to match one side of the connection and IP filter is used to match the other side?

msherif1234: the idea here is as follows
- on ingress side
we match on the src CIDR if known if its not we can use `0.0.0.0/0` then u have number of options to further narrow down one of them the dstIP
- on the egress since the cluster is the one generating the traffic we filter on dst CIDR again if we don't know we can use `0.0.0.0/0` and further filter on srcIP that will become very handy when use feature like OVN EgressIP where u want to see all flows with srcIP == EgressIP

and yes this feature need doc to explain all options

msherif1234: cc'd @skrthomas

skrthomas: Thanks @msherif1234 just want to make sure I'm tying this in the right place, is this part of the netobserv CLI?

jotak: > and yes this feature need doc to explain all options
I mean, the doc here in this repo must be clarified:

```go
	// FlowFilterIPCIDR is the IP CIDR to filter flows.
	// Example: 10.10.10.0/24 or 100:100:100:100::/64
	FlowFilterIPCIDR string `env:"FLOW_FILTER_IP_CIDR"`

	// FlowFilterIP is the IP to filter flows.
	// Example: 10.10.10.10
	FlowFilterIP string `env:"FLOW_FILTER_IP"`
```

=> it tells nothing about the fact that CIDR is used to filter on the remote peer (src for ingress and dst for egress), and IP filter for the local end (dst for ingress and src for egress). That's to be documented here as well.

Also there's a config.md file to update iirc

jotak: I'm still unsure about src/dst decision that is made here regarding direction. I haven't tested but I think that's going to miss some flows. I can try testing when I find some time

msherif1234: I added https://github.com/netobserv/netobserv-ebpf-agent/pull/307/commits/f2a65b3639d01c06092555312c2b85b43881aa5d to explain in details the feature and some possible use cases

docs/flow_filtering.md
```diff
@@ -0,0 +1,95 @@
...
+Flow rule-base filtering is a method to control the flow of packets cached in eBPF's flows table based on certain configuration
+
+## Flow filter rule configuration
+
+ The Flow filter rule consists of two parts mandatory and optional parameters.
+ 
+### Mandatory parameters
+
+- `FLOW_FILTER_IP_CIDR` - IP address and CIDR mask for the flow filter rule, supports IPv4 and IPv6 address format.
+  If wanted to match against any IP, user can use `0.0.0.0/0` or `::/0` for IPv4 and IPv6 respectively.
```
memodi, line 14: IMO `0.0.0.0/0` or `::/0` is not a good example, where you're accepting everything (current state) or rejecting everything, specific example of specific CIDR/IP address would be helpful.

docs/flow_filtering.md
```diff
@@ -0,0 +1,95 @@
...
+- `FLOW_FILTER_ACTION` - Action to be taken for the flow filter rule. Possible values are `Accept` and `Reject`.
+  - For the matching rule with `Accept` action this flow will be allowed to be cached in eBPF table, with updated global metric `FlowFilterAcceptCounter`.
+  - For the matching rule with `Reject` action this flow will be dropped and not cached in eBPF table, with updated global metric `FlowFilterRejectCounter`.
+  - If the rule is not matched, the flow will be allowed to be cached in eBPF table, with updated global metric `FlowFilterNoMatchCounter`.
```
memodi, line 18: this is also implicit `Accept`, no? I realize `FlowFilterAcceptCounter` and `FlowFilterRejectCounter` are specific if filter is matched, as opposed to having a metric, I wonder if there a better way to flag when the filter is matched vs not matched may be propagating to our consolePlugin ?

docs/flow_filtering.md
```diff
@@ -0,0 +1,95 @@
...
+- When filtering egress traffic, the destination IP address is a crucial parameter used to match filter rules configured with CIDR notation.
+  CIDR notation allows for the efficient representation of IP address ranges by combining the base IP address with a prefix length.
+
+### Matching specific endpoints using `FLOW_FILTER_IP`
+
+- When filtering ingress traffic, the `FLOW_FILTER_IP` will be used to match against the destination IP address of the packet, 
+  so after matching source IP CIDR now we can pinpoint to a specific endpoint via `FLOW_FILTER_IP` and based on the provisioned action,
+  flow with either be allowed to be cached in eBPF flow table or not.
+- When filtering egress traffic, the `FLOW_FILTER_IP` will be used to match against the source IP address of the packet,
+  so after matching destination IP CIDR now we can pinpoint to a specific endpoint via `FLOW_FILTER_IP` and based on the provisioned action,
```
memodi, line 48: IMO the config names are quite confusing to represent what they're doing and I do need to remember what they'll match against depending upon the direction or would have to consult API doc, is it possible to make it more simpler? 
Also, `FLOW_FILTER_IP` doesn't accept CIDR, IIUC usually the cluster resources belong to fix CIDR space , it would be good to be able specify CIDR space for this config as well. I may not be aware any restriction/limitation that led to have specific IP instead of CIDR space.

docs/flow_filtering.md
```diff
@@ -0,0 +1,95 @@
...
+  - For the matching rule with `Accept` action this flow will be allowed to be cached in eBPF table, with updated global metric `FlowFilterAcceptCounter`.
+  - For the matching rule with `Reject` action this flow will be dropped and not cached in eBPF table, with updated global metric `FlowFilterRejectCounter`.
+  - If the rule is not matched, the flow will be allowed to be cached in eBPF table, with updated global metric `FlowFilterNoMatchCounter`.
+
+### Optional parameters
+
+- `FLOW_FILTER_DIRECTION` - Direction of the flow filter rule. Possible values are `Ingress` and `Egress`.
+- `FLOW_FILTER_PROTOCOL` - Protocol of the flow filter rule. Possible values are `TCP`, `UDP`, `SCTP`, `ICMP`, `ICMPv6`.
+- `FLOW_FILTER_SOURCE_PORT` - Single Source port of the flow filter rule.
+- `FLOW_FILTER_SOURCE_PORT_RANGE` - Source port range of the flow filter rule. using "80-100" format.
```
memodi, line 25: assuming port and port-range are mutually exclusive?

docs/flow_filtering.md
memodi, line 48: would it be correct to say `FLOW_FILTER_IP` will match the in-cluster IP address whether it's ingress or egress?

msherif1234: yes

docs/flow_filtering.md
msherif1234: yes either one port or range but not both

docs/flow_filtering.md
```diff
@@ -0,0 +1,95 @@
...
+
+Flow rule-base filtering is a method to control the flow of packets cached in eBPF's flows table based on certain configuration
+
+## Flow filter rule configuration
+
+ The Flow filter rule consists of two parts mandatory and optional parameters.
+ 
+### Mandatory parameters
+
+- `FLOW_FILTER_IP_CIDR` - IP address and CIDR mask for the flow filter rule, supports IPv4 and IPv6 address format.
```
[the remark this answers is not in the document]
msherif1234: that was explained below in the doc

docs/flow_filtering.md
msherif1234: that doesn't mean u are accepting see the fine tune section u can use the different optional configs to narrow it down for example I wanted to see all flows with `sport 80` regardless where it came from or going to

docs/flow_filtering.md
msherif1234: we have 3 counters one for accept action and another for reject and the last one for nomatch flows
can u explain why nomatch metrics isn't good enough also this can be used with/without console

docs/flow_filtering.md
msherif1234: not sure why u need CIDRs for in cluster address as normally this is very specific IP?

memodi: For `FLOW_FILTER_IP` config which would match in-cluster IPs, I was thinking about the use cases where I am interested in to see flows for multiple pods/service in a cluster NS to filter-in the flowlogs for.

pkg/ebpf/flow_filter.go
```diff
@@ -0,0 +1,201 @@
...
+		val.Direction = BpfDirectionTMAX_DIRECTION
+	}
+
+	switch config.FlowFilterAction {
+	case "Reject":
+		val.Action = BpfFilterActionTREJECT
+	case "Accept":
+		val.Action = BpfFilterActionTACCEPT
+	default:
+		val.Action = BpfFilterActionTMAX_FILTER_ACTIONS
```
jotak, line 98: since this is a required parameter, shouldn't we fail here and avoid having to deal with invalid config in the bpf program?

msherif1234: the ebpf code has handle for this special value to just ignore filtering all together plus if we provision from the operator taht won't happen

jotak (approved):
lgtm

I also tested a use-case such as "filter all traffic (in and out) for a given pod"; then compared the results without ebpf filter but doing that same filter from the UI to check we get roughly the same flows & traffic
=> works as expected!

docs/flow_filtering.md
memodi: it's fine to have NoMatch filter, I just thought NoMatch would be equivalent of "Accept" but I see your point with help of this metric we could identify the filter is being applied or not. thanks!

pkg/metrics/metrics.go
```diff
@@ -85,6 +85,13 @@ var (
 		"source",
 		"reason",
 	)
+	filterFlows = defineMetric(
+		"filtered_flows_total",
+		"Number of filtered flows",
+		TypeCounter,
+		"source",
+		"reason",
+	)
```
jpinsonneau, line 94: Do we need to update the dashboards ?

Maybe we can manage that in https://github.com/netobserv/network-observability-operator/pull/607 or create a followup in Jira

bpf/flows_filter.h
```diff
@@ -0,0 +1,240 @@
...
+                    }
+                    if (rule->icmpCode != 0) {
+                        if (rule->icmpCode == id->icmp_code) {
+                            BPF_PRINTK("icmpCode matched\n");
+                            result++;
+                        } else {
+                            result = 0;
+                            goto end;
+                        }
+                    }
```
jpinsonneau, line 136: I don't think it makes sense to have `icmpCode` independent of `icmpType` so we should move that check inside of `if (rule->icmpType != 0)` condition above and document that `icmpCode` requires `icmpType` to be set.

WDYT ?

msherif1234: they  are two independent filters so u set one or the other or both or none

pkg/metrics/metrics.go
msherif1234: not sure if we need to update the dashboard or we just add it to the query list not sure @jotak but that is correct we can either added to 607 PR on a following PR

jotak: let's do a follow-up JIRA then; I would say the goal for NETOBSERV-1478 is reached already

For users who enable filtering, I guess it can be interesting to know the ratio of filtered in/filtered out flows indeed, so yeah why not adding that to the health dashboard

bpf/flows_filter.h
jpinsonneau: yes, but each type have its own related code implementation:
https://www.iana.org/assignments/icmp-parameters/icmp-parameters.xhtml

so it doesn't make any sense to have `icmpCode` filter only

jpinsonneau (approved):
LGTM, thanks @msherif1234

PR #341, opened 2024-05-27
Codecov: remove token, turn off CI failure

jotak:
See also: netobserv/network-observability-console-plugin#531

Also, align the workflow file with other repos

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,10 +1,8 @@
 name: Pull request checks
 
 on:
-  push:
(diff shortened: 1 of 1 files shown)


.github/workflows/pull_request.yml
```diff
@@ -1,10 +1,8 @@
 name: Pull request checks
 
 on:
-  push:
```
jotak, line 4: no need to have that on 'push' events: we already run that on 'push' from the other workflow file, `push_image.yaml`

PR #348, opened 2024-06-14
NETOBSERV-1564: do not force flushing maps when rb is triggered

jotak:
## Description

Flushing (without throttling) has a nefast effect in high stressed scenario, generating a lot of evictions from maps, resulting in many more flows generated.

High stressed scenarios should rather rely on rb+accounter, which better handles the number of generated flows.

Also, use errno as the reason for the metric

With this change + high stress scenario I'm seeing better CPU but more memory slightly increased:

Capture d’écran du 2024-06-14 17-14-26

overall, -50% CPU and +10% memory

This should be tested against cluster-density-v2

## Dependencies

n/a

## Checklist


pkg/flow/tracer_ringbuf.go
```diff
@@ -91,15 +91,9 @@ func (m *RingBufTracer) listenAndForwardRingBuffer(debugging bool, forwardCh cha
 	if debugging {
 		m.stats.logRingBufferFlows(mapFullError)
 	}
-	// if the flow was received due to lack of space in the eBPF map
```
msherif1234, line 94: wouldn't removing this logic means we will stay in hmap full condition now for longer and possibly dropping flows ?

jotak: I agree that having a fixed-rate flush could be a better solution, but actually there's already a flush every 5s (the `cacheActiveTimeout` setting) ; users can set it to 1s if that works better for them

jotak: or we introduce a new "stress timer" (could be automatically set to `cacheActiveTimeout / something`) that only starts when stress is detected (ie. maps are full) ? but that's a bit complex, it has to be worth it

jotak: by the way: when you said "this logic means we will stay in hmap full condition now for longer and possibly dropping flows" I don't think this is true; when maps are full, flows are not dropped, they're moved to the ring buffer. Dropped flows are only for busy maps when there isn't a new flow creation, so not something that can lead the map being full (we're updating an entry in that case, not adding a new one, so no incidence on hmap size)

msherif1234: we use rb for new flows and it send 1 pkt at a time to userspace so it can't handle burst of traffic either

jotak: it still handle that much better than an uncontrolled hot-loop of flush events. Look at the -50% CPU that I'm showing in the PR description

jotak: I'll run perf-scale tests to check the diff

msherif1234: /LGTM if perf scale shows improvements

jotak: ingress-perf shows slightly improved stats: -4% memory, -8% CPU
(https://docs.google.com/spreadsheets/d/1EN12dogz-0_H_5tSoV24T4iHa9a3wSgoF396YyOqavA/edit?gid=93930639#gid=93930639 / diff: https://docs.google.com/spreadsheets/d/1q_XCJ48h2Q78JapxgNB37nJe-4lTYVlMx7xVrn7ztck/edit?gid=696044201#gid=696044201)

Now running cluster-density. But I'm not sure to see something very different here, as the RB isn't involved a lot anyway.

Problem is that, none of those test really stress the agents. So I may not see something better than this -8% here. The tests that I did above DO stress the agents much more, and I had -50%

jotak: @msherif1234 , @Amoghrd did run cluster-density, with also slightly positive results: https://issues.redhat.com/browse/NETOBSERV-1564?focusedId=24995474&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-24995474

jotak:
thanks @msherif1234 
/approve

PR #360, opened 2024-06-27
NETOBSERV-1732: add logic to lkup all previous tc filters and remove them

msherif1234:
## Description

Today if ebpf agent pod is killed with `SIGKILL` the clean up code never get a chance to remove all TC filters and qdisc installed 
this PR check all previously installed TC filters and remove them

## Dependencies

n/a

## Checklist

### DIFF
--- a/pkg/ebpf/tracer.go
+++ b/pkg/ebpf/tracer.go
@@ -23,6 +23,7 @@ import (
 	"github.com/vishvananda/netlink"
 	"github.com/vishvananda/netns"
 	"golang.org/x/sys/unix"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
 )
 
 // $BPF_CLANG and $BPF_CFLAGS are set by the Makefile.
@@ -42,6 +43,8 @@ const (
 	pktDropHook              = "kfree_skb"
 	constPcaEnable           = "enable_pca"
 	pcaRecordsMap            = "packet_record"
+	tcEgressFilterName       = "tc/tc_egress_flow_parse"
+	tcIngressFilterName      = "tc/tc_ingress_flow_parse"
 )
 
 var log = logrus.WithField("component", "ebpf.FlowFetcher")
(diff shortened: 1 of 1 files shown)


jotak:
@msherif1234 I thought it was already managed e.g. here: https://github.com/netobserv/netobserv-ebpf-agent/blob/main/pkg/ebpf/tracer.go#L308 , what's the difference exactly with the existing deletions?

msherif1234:
> @msherif1234 I thought it was already managed e.g. here: https://github.com/netobserv/netobserv-ebpf-agent/blob/main/pkg/ebpf/tracer.go#L308 , what's the difference exactly with the existing deletions?
> 
> Is it because previously it would only delete for the current interfaces, but not if interfaces have changed in the meantime?

current logic won't work as when the pod restart the FD value won't be the same as old program so the existing code won't ever kick in https://github.com/netobserv/netobserv-ebpf-agent/blob/main/pkg/ebpf/tracer.go#L304

pkg/ebpf/tracer.go
```diff
@@ -242,6 +244,86 @@ func (m *FlowFetcher) AttachTCX(iface ifaces.Interface) error {
...
+		return err
+	}
+
+	filters, err := netlink.FilterList(link, tcDir)
+	if err != nil {
+		return err
+	}
+
+	for _, f := range filters {
+		if err := netlink.FilterDel(f); err != nil {
```
jotak, line 259: not sure how likely it is to get an error here, but to be more resilient you could catch all errors in a slice and use [errors.Join](https://pkg.go.dev/errors#Join) : that would avoid returning after an error in case there's still filters to delete in the list.

pkg/ebpf/tracer.go
```diff
@@ -242,6 +244,86 @@ func (m *FlowFetcher) AttachTCX(iface ifaces.Interface) error {
...
+		if err != nil {
+			return fmt.Errorf("listing egress filters: %w", err)
+		}
+		for _, filter := range egressFilters {
+			if bpfFilter, ok := filter.(*netlink.BpfFilter); ok {
+				if strings.HasPrefix(bpfFilter.Name, tcEgressFilterName) {
+					{
+						egressDevs = append(egressDevs, l)
+					}
+				}
```
jotak, line 303: Nit: unnecessary `{}`

```suggestion
				if strings.HasPrefix(bpfFilter.Name, tcEgressFilterName) {
					egressDevs = append(egressDevs, l)
				}
```

jotak (commented):
some minor comments, lgtm otherwise

pkg/ebpf/tracer.go
msherif1234: I can use Aggregrate error too which does the same like errors join

jotak (approved):
/lgtm

PR #389, opened 2024-08-15
NETOBSERV-1803: Allow flow filtering for L4 protocols using two ports

msherif1234:
## Description
For example allow TCP ports="80,100"

## Dependencies

n/a

## Checklist


pkg/ebpf/flow_filter.go
```diff
@@ -163,43 +172,67 @@ func (f *Filter) getFilterValue(config *FilterConfig) (BpfFilterValueT, error) {
 	return val, nil
 }
 
-func getSrcPorts(config *FilterConfig) (uint16, uint16) {
+func getSrcPortsRange(config *FilterConfig) (uint16, uint16) {
 	if config.FilterSourcePort.Type == intstr.Int {
 		return uint16(config.FilterSourcePort.IntVal), 0
 	}
-	start, end, err := getPortsFromString(config.FilterSourcePort.String())
+	start, end, err := getPortsFromString(config.FilterSourcePort.String(), "-")
```
stleerh, line 179: If only the start value is given (int), then make the start and end value be the same.

```
    return uint16(config.FilterSourcePort.IntVal), uint16(config.FilterSourcePort.IntVal)
```

The code in flows\_filter.h would be simplified by not having to check whether one or both values are set since either none or both are set.

pkg/ebpf/flow_filter.go
```diff
@@ -163,43 +172,67 @@ func (f *Filter) getFilterValue(config *FilterConfig) (BpfFilterValueT, error) {
...
 }
 
-func getDstPorts(config *FilterConfig) (uint16, uint16) {
+func getSrcPorts(config *FilterConfig) (uint16, uint16) {
+	port1, port2, err := getPortsFromString(config.FilterSourcePort.String(), ",")
+	if err != nil {
+		return 0, 0
+	}
+	return port1, port2
+}
```
stleerh, line 192: To support more than two ports, instead of port1 and port2, return a port array.  Then it just needs a C function in flow\_filter.h to check whether a port is in the array or not.  The 'if' statement will even be shorter.

msherif1234: the prob is not about arry of not the main prob is we configure the agent via envvar and we cant use array there unless we do something else instead of envvar

pkg/ebpf/flow_filter.go
msherif1234: the bpf code check if start only set to know if its single value but that could change bit what is the benefit ?  bpf code use start end only for range I am not sure that wll be simplified ?

msherif1234: let me think more about this I am all in for simplification

msherif1234: then can be done if we ensure `x,y` fmt where x < y which might not too bad wdyt ?

msherif1234: this won't work as when use `53,5353` u need to look only for ports matching those 2 values we can't use start and end as taht will bring in non interesting ports in the middle

pkg/ebpf/flow_filter.go
stleerh: Why wouldn't you be able to set the environment variable, FILTER_SOURCE_PORTS, to contain more than two ports like "53,5353,80,8000" or any number of individual ports?

pkg/ebpf/flow_filter.go
stleerh: This was just a simplification for port range, not for individual ports, for the logic you had to have on lines 63 and 84 in flow_filter.h.  It's minor.

pkg/ebpf/flow_filter.go
msherif1234: we can't use variable sized arrays so we have to have known size how big that array will be

stleerh: okay, that's quite limiting.

bpf/types.h
```diff
@@ -212,10 +212,16 @@ struct filter_value_t {
...
+    u16 dstPort1;
+    u16 dstPort2;
     u16 srcPortStart;
     u16 srcPortEnd;
+    u16 srcPort1;
+    u16 srcPort2;
     u16 portStart;
     u16 portEnd;
+    u16 port1;
+    u16 port2;
```
jpinsonneau, line 224: Why not having more than two values here ? 
Is it just a matter of keeping the code simple here or are there other limitations ?

jpinsonneau (approved):

bpf/types.h
msherif1234: no we can have more but the question is how many more ? it can't be variable size or dynamic so I just limited to two for now per rule

jpinsonneau: that's something we should define according to the use cases we want to cover :smile_cat: 

cc @stleerh @jotak

jpinsonneau (approved):

bpf/types.h
jotak: I guess if someone wants a third one, we can move to a fixed size array such as 8 ? (next time)

PR #438, opened 2024-10-24
NETOBSERV-1965: eBPF packet translation tracker

msherif1234:
## Description

Ability to detect packet transformation and use it to enrich existing flows
## Dependencies
n/a

## Checklist


msherif1234:
image

pkg/agent/config.go
```diff
@@ -235,6 +235,8 @@ type Config struct {
 	EnableNetworkEventsMonitoring bool `env:"ENABLE_NETWORK_EVENTS_MONITORING" envDefault:"false"`
 	// NetworkEventsMonitoringGroupID to allow ebpf hook to process samples for specific groupID and ignore the rest
 	NetworkEventsMonitoringGroupID int `env:"NETWORK_EVENTS_MONITORING_GROUP_ID" envDefault:"10"`
+	// EnablePktTransformationTracking allow tracking packets after transformation for example NAT, default is false.
```
jotak, line 238: is it a common naming to talk about "packet transformation" ? I had the feeling it was more "translation" ; but I'm not very well aware of that

pkg/model/flow_content.go
```diff
@@ -84,12 +84,17 @@ func (p *BpfFlowContent) AccumulateAdditional(other *ebpf.BpfAdditionalMetrics)
...
 		p.AdditionalMetrics.FlowRtt = other.FlowRtt
 	}
+	// Network events
 	for _, md := range other.NetworkEvents {
 		if !AllZerosMetaData(md) && !networkEventsMDExist(p.AdditionalMetrics.NetworkEvents, md) {
 			copy(p.AdditionalMetrics.NetworkEvents[p.AdditionalMetrics.NetworkEventsIdx][:], md[:])
 			p.AdditionalMetrics.NetworkEventsIdx = (p.AdditionalMetrics.NetworkEventsIdx + 1) % MaxNetworkEvents
 		}
 	}
+	// Packet Translations
```
jotak, line 94: it would be interesting to know if there's information potentially lost here. Eg. if two map/cpus report two different things for the same flow. idk if that's possible.

jotak: also, do you know if zoneid can be 0 (other than when it's unset) ? If so, might be more efficient to check for existing zoneid, than checking for non-zero IPs

pkg/pbflow/proto.go
```diff
@@ -98,9 +104,17 @@ func FlowToPB(fr *model.Record, s *ovnobserv.SampleDecoder) *Record {
 	if fr.Metrics.EthProtocol == model.IPv6Type {
 		pbflowRecord.Network.SrcAddr = &IP{IpFamily: &IP_Ipv6{Ipv6: fr.ID.SrcIp[:]}}
 		pbflowRecord.Network.DstAddr = &IP{IpFamily: &IP_Ipv6{Ipv6: fr.ID.DstIp[:]}}
+		if fr.Metrics.AdditionalMetrics != nil {
```
jotak, line 107: in dual stack is it possible to have traffic over ipv4 NATed to ipv6, or the other way around? In other words, are we sure that the EthProtocol that we got in the first place is still the same in xlat ?
If not, then we don't know here if the xlat addresses are v4 or v6 ?

pkg/model/flow_content.go
msherif1234: yes zoneid of 0 is valid

pkg/agent/config.go
msherif1234: I have seen both name used alternatively

pkg/pbflow/proto.go
msherif1234: in theory its possible but I don't think k8s service will change from v4 to v6 but we can fix that when this happen ?

jotak:
I'll write my few concerns in a new JIRA (2 things: fixing the dual-stack issue, and being more consistent on naming if we can use the same terminology everywhere ie. Translate instead of Transformation - and the latter seems less commonly used to me)
But it's ok to merge to unblock the queue
/lgtm

jotak:
Thanks
/lgtm

PR #49, opened 2022-08-29
Simplify eBPF code optimizations

mariomac:
This PR simplifies some parts of the great PR from Pravein.

Please see inline comments for more details/explanations about my changes.

PR #444, opened 2024-11-04
NETOBSERV-1954: fix under-estimation of traffic

jotak:
Remove flawed logic that ignores some flows, especially visible under stress
This came from that time: https://github.com/netobserv/netobserv-ebpf-agent/pull/49/files#diff-f3248f855cc319b6d96125842a513586d6d8cbfb1e241a93e79911d08e41c728L334-L336

Old comment said: "...when a new flow maps to the same entry, it has some zombie entries." - this is weird and I'm not sure how that would be possible, flow_metrics entries should be zeroed as we're using designed initializers. In any case, relying on a single eviction time reference is wrong because flows are evicted one by one, resulting in many new flows having their timer set before the "last eviction time".

If we find evidence of those "zombie entries" then we must find another way to fix them.

## Dependencies

n/a

## Checklist

* [x] Is this PR backed with a JIRA ticket? If so, make sure it is written as a title prefix _(in general, PRs affecting the NetObserv/Network Observability product should be backed with a JIRA ticket - especially if they bring user facing changes)._

### DIFF
--- a/pkg/flow/tracer_map.go
+++ b/pkg/flow/tracer_map.go
@@ -26,7 +26,6 @@ type MapTracer struct {
 	staleEntriesEvictTimeout time.Duration
 	// manages the access to the eviction routines, avoiding two evictions happening at the same time
 	evictionCond               *sync.Cond
-	lastEvictionNs             uint64
 	metrics                    *metrics.Metrics
 	timeSpentinLookupAndDelete prometheus.Histogram
 }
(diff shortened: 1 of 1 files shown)


pkg/model/record.go
```diff
@@ -92,7 +92,7 @@ func NewRecord(
 
 func Accumulate(r *ebpf.BpfFlowMetrics, src *ebpf.BpfFlowMetrics) {
 	// time == 0 if the value has not been yet set
-	if r.StartMonoTimeTs == 0 || r.StartMonoTimeTs > src.StartMonoTimeTs {
+	if r.StartMonoTimeTs == 0 || (r.StartMonoTimeTs > src.StartMonoTimeTs && src.StartMonoTimeTs != 0) {
```
msherif1234, line 95: isn't the `&&` logic is redundant ?

jotak: no, first is "r.StartMonoTimeTs", second is "src.StartMonoTimeTs"

jotak:
thanks @msherif1234 , I'm merging this and I'll also pay close attention to the perf tests wrt observed traffic 
/approve

PR #457, opened 2024-11-13
Adapt to ovnk-obs change

jotak:
## Dependencies

n/a


jotak:
@msherif1234 that's a first step to just bump ovnk without actually adding the structured data to the pipeline
next step is to send structure data instead of string

PR #458, opened 2024-11-13
NETOBSERV-1974: Refactor network events to return list of maps instead of strings

msherif1234:
## Description

Network events was orignally decode as string with this recent ovnk lib it will come as struct and the agent need to generate list of configmaps with those events
this effort to allow generating metrics with network events

## Dependencies

https://github.com/ovn-kubernetes/ovn-kubernetes/pull/4818
## Checklist


pkg/pbflow/proto.go
```diff
@@ -99,13 +101,32 @@ func FlowToPB(fr *model.Record, s *ovnobserv.SampleDecoder) *Record {
...
+								Events: map[string]string{"Name": e.Name},
+							},
+							{
+								Events: map[string]string{"Namespace": e.Namespace},
+							},
+							{
+								Events: map[string]string{"Direction": e.Direction},
+							},
+						}
+						pbflowRecord.NetworkEventsMetadata = append(pbflowRecord.NetworkEventsMetadata, pbEvents...)
```
jotak, line 129: This shouldn't be one map per single event field, but one map per event (with all event fields in the same map)

```suggestion
						e := ovnmodel.ACLEvent{NetworkEvent: md}
						pbEvent := NetworkEvent{
								Events: map[string]string{
								    "Action": e.Action,
								    "Feature": e.Actor,
								    "Name": e.Name,
							            "Namespace": e.Namespace,
								    "Direction": e.Direction,
							},
						}
						pbflowRecord.NetworkEventsMetadata = append(pbflowRecord.NetworkEventsMetadata, pbEvents)
```

right?

pkg/pbflow/proto.go
```diff
@@ -99,13 +101,32 @@ func FlowToPB(fr *model.Record, s *ovnobserv.SampleDecoder) *Record {
...
-		seen := make(map[string]bool)
+		seen := make(map[ovnmodel.NetworkEvent]bool)
+		pbflowRecord.NetworkEventsMetadata = make([]*NetworkEvent, 0)
 		for _, metadata := range fr.Metrics.NetworkEvents {
 			if !model.AllZerosMetaData(metadata) {
 				if md, err := s.DecodeCookie8Bytes(metadata); err == nil {
 					protoLog.Debugf("Network Events Metadata %v decoded Cookie: %v", metadata, md)
 					if !seen[md] {
-						pbflowRecord.NetworkEventsMetadata = append(pbflowRecord.NetworkEventsMetadata, md)
+						e := ovnmodel.ACLEvent{NetworkEvent: md}
```
jotak, line 111: Also you don't have the guarantee that md is of type NetworkEvent
Should be something like:
```go

					if acl, ok := md.(*ovnmodel.ACLEvent); ok {
						// (fill pbEvent)
					} else {
                                               // something generic here, such as:
                                               pbEvent := NetworkEvent{
								Events: map[string]string{
								    "Message": md.String()
                                                                },
                                                          },
					}
```

pkg/pbflow/proto.go
msherif1234: yes I addressed this change

pkg/pbflow/proto.go
msherif1234: I don't think we will hit a case that will generate string but can't get cast as ACLEvent but that change won't hurt

pkg/pbflow/proto.go
```diff
@@ -100,13 +102,34 @@ func FlowToPB(fr *model.Record, s *ovnobserv.SampleDecoder) *Record {
...
+		pbflowRecord.NetworkEventsMetadata = make([]*NetworkEvent, 0)
 		for _, metadata := range fr.Metrics.NetworkEvents {
+			var pbEvent NetworkEvent
 			if !model.AllZerosMetaData(metadata) {
 				if md, err := s.DecodeCookie8Bytes(metadata); err == nil {
 					protoLog.Debugf("Network Events Metadata %v decoded Cookie: %v", metadata, md)
-					if !seen[md] {
-						pbflowRecord.NetworkEventsMetadata = append(pbflowRecord.NetworkEventsMetadata, md)
-						seen[md] = true
+					acl, ok := md.(*ovnmodel.ACLEvent)
```
jotak, line 111: md.String() is called three times below I guess it could be set in a var instead

jotak (approved):
LGTM

PR #466, opened 2024-11-25
NETOBSERV-1994: remove unneeded bpf map update calls

msherif1234:
## Description

bpf code was doing unnecessary calls to `bpf_map_update_elem` while the map is already been updated in place. which also added unneeded cpu load 
## Dependencies

n/a

## Checklist


bpf/flows.c
```diff
@@ -152,7 +139,8 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
             if (trace_messages) {
                 bpf_printk("error adding flow %d\n", ret);
             }
-
+            // Update global counter for hashmap update errors
+            increase_counter(HASHMAP_FLOWS_DROPPED);
```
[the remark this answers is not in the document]
msherif1234: ok yeah wanted to know when hmap not used and we do rb but I agree its duplicate I will remove it

jotak (approved):
LGTM

bpf/dns_tracker.h
```diff
@@ -92,7 +92,7 @@ static __always_inline int track_dns_packet(struct __sk_buff *skb, pkt_info *pkt
         if ((flags & DNS_QR_FLAG) == 0) { /* dns query */
             fill_dns_id(pkt->id, &dns_req, dns_id, false);
             if (bpf_map_lookup_elem(&dns_flows, &dns_req) == NULL) {
-                bpf_map_update_elem(&dns_flows, &dns_req, &ts, BPF_ANY);
+                bpf_map_update_elem(&dns_flows, &dns_req, &ts, BPF_NOEXIST);
```
tohojo, line 95: You can skip the lookup before this update when using BPF_NOEXIST :)

bpf/flows.c
```diff
@@ -142,7 +129,7 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
 
         // even if we know that the entry is new, another CPU might be concurrently inserting a flow
         // so we need to specify BPF_ANY
```
tohojo, line 131: This comment is wrong now

tohojo (commented):
A few nits, but also a more general comment: you're not handling failure of bpf_map_update(...,BPF_NOEXIST), which means that if two threads end up doing this concurrently, one of the creation attempts will be lost.

This is no worse than before the patch, though (before the patch, one attempt would silently overwrite the other, now one will just fail). But, well, it's possible to do better, so I would suggest handling the errors where possible :)

bpf/flows.c
msherif1234: removed

bpf/dns_tracker.h
msherif1234: done

msherif1234:
> A few nits, but also a more general comment: you're not handling failure of bpf_map_update(...,BPF_NOEXIST), which means that if two threads end up doing this concurrently, one of the creation attempts will be lost.
> 
> This is no worse than before the patch, though (before the patch, one attempt would silently overwrite the other, now one will just fail). But, well, it's possible to do better, so I would suggest handling the errors where possible :)

are u referring to DNS update or this is more of general comment and u need better handling for map update errors ? can u expand more if that is the case ?

bpf/flows.c
```diff
@@ -140,31 +133,36 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
...
-                (flow_record *)bpf_ringbuf_reserve(&direct_flows, sizeof(flow_record), 0);
-            if (!record) {
-                if (trace_messages) {
-                    bpf_printk("couldn't reserve space in the ringbuf. Dropping flow");
+            if (ret == -EEXIST) {
+                flow_metrics *aggregate_flow =
+                    (flow_metrics *)bpf_map_lookup_elem(&aggregated_flows, &id);
+                if (aggregate_flow != NULL) {
+                    update_existing_flow(aggregate_flow, &pkt, dns_errno, len);
+                }
```
tohojo, line 146: You could keep the HASHMAP_FLOWS_DROPPED counter and update it in an "else" branch here. I would expect this to basically never happen, but just to be on the safe side, and since you already have that counter.

bpf/flows.c
```diff
@@ -140,31 +133,36 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
...
+        long ret = bpf_map_update_elem(&aggregated_flows, &id, &new_flow, BPF_NOEXIST);
         if (ret != 0) {
-            // usually error -16 (-EBUSY) or -7 (E2BIG) is printed here.
-            // In this case, we send the single-packet flow via ringbuffer as in the worst case we can have
-            // a repeated INTERSECTION of flows (different flows aggregating different packets),
-            // which can be re-aggregated at userpace.
-            // other possible values https://chromium.googlesource.com/chromiumos/docs/+/master/constants/errnos.md
             if (trace_messages) {
                 bpf_printk("error adding flow %d\n", ret);
             }
```
tohojo, line 140: I would not personally consider EEXIST an error that needs logging (here, or in other places you have trace logging). Only if the subsequent lookup then fails to return an entry (see comment below).

bpf/flows.c
msherif1234: yeah that is why I felt its less useful but I can reuse it when eexist fail to lkup

bpf/flows.c
msherif1234: the thing is map update can fail for many reasons ebusy or e2big so i need to capture those as well , now if eexist error will be happening alot I can filter out the trace for exist error and with counter back we will know when we drop flows

tohojo: Yeah, so that's basically what I meant: Use the trace message only for errors other than EEXIST (just move the log statement a bit further down where you're handling those anyway)

jotak (approved):
LGTM
I tested the first version which worked well, will need to retest when you tell me it's stable

bpf/flows.c
```diff
@@ -93,33 +112,7 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
...
-            // a duplicated UNION of flows (two different flows with partial aggregation of the same packets),
-            // which can't be deduplicated.
-            // other possible values https://chromium.googlesource.com/chromiumos/docs/+/master/constants/errnos.md
-            if (trace_messages) {
-                bpf_printk("error updating flow %d\n", ret);
-            }
-            // Update global counter for hashmap update errors
-            increase_counter(HASHMAP_FLOWS_DROPPED);
-        }
+        update_existing_flow(aggregate_flow, &pkt, dns_errno, len);
```
jotak, line 115: must swap len and dns_errno

```suggestion
        update_existing_flow(aggregate_flow, &pkt, len, dns_errno);
```

bpf/flows.c
```diff
@@ -140,31 +133,42 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
...
-            flow_record *record =
-                (flow_record *)bpf_ringbuf_reserve(&direct_flows, sizeof(flow_record), 0);
-            if (!record) {
-                if (trace_messages) {
-                    bpf_printk("couldn't reserve space in the ringbuf. Dropping flow");
+            if (ret == -EEXIST) {
+                flow_metrics *aggregate_flow =
+                    (flow_metrics *)bpf_map_lookup_elem(&aggregated_flows, &id);
+                if (aggregate_flow != NULL) {
+                    update_existing_flow(aggregate_flow, &pkt, dns_errno, len);
```
jotak, line 145: here as well 
```suggestion
                    update_existing_flow(aggregate_flow, &pkt, len, dns_errno);
```

jotak (approved):

Amoghrd:
@msherif1234 Ran full automated suite against the PR and it passed(barring some flakes)
[no-regression#9](https://mastern-jenkins-csb-openshift-qe.apps.ocp-c1.prod.psi.redhat.com/job/netobserv/job/no-regression/9/) 
/label qe-approved

PR #469, opened 2024-11-27
NETOBSERV-1995: Use global map, spinlock, split maps

jotak:
Also:
- remove MAC and eth_protocol from map key (they're moved to flow
  metrics ie. map values)
- in every hook, we don't update flow counters, even when the flow doesn't exist and is created. Previously, we were
adding bytes/packet counters from hooks, which doesn't seem correct as
counting bytes/packets is the role of the TC hook only - else it can
lead to overcounts I guess
- remove code that sets MIN_RTT on TCP flows (not sure why we had that)
- small refactoring of the "Accumulate" functions to adapt to new model


jotak:
Perf tests (https://docs.google.com/spreadsheets/d/1qakBaK1dk_rERO30k1cSR4W-Nn0SXW4A3lqQ1sZC4rE/edit?gid=48358889#gid=48358889) show ~ -25% CPU usage with the global map

bpf/flows.c
```diff
@@ -91,58 +91,48 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
...
+        aggregate_flow->flow_metrics.flags |= pkt.flags;
+        aggregate_flow->flow_metrics.dscp = pkt.dscp;
+        aggregate_flow->flow_metrics.dns_record.id = pkt.dns_id;
+        aggregate_flow->flow_metrics.dns_record.flags = pkt.dns_flags;
+        aggregate_flow->flow_metrics.dns_record.latency = pkt.dns_latency;
+        aggregate_flow->flow_metrics.dns_record.errno = dns_errno;
+        bpf_spin_unlock(&aggregate_flow->lock);
     } else {
         // Key does not exist in the map, and will need to create a new entry.
-        u64 rtt = 0;
```
jotak, line 125: @msherif1234 I don't understand the rationale for this code. We're setting default RTT for TCP flows, how could that be accurate? I don't see how this wouldn't produce fake data.

jotak:
cc @tohojo

jotak:
removing packed attribute gives another 20% CPU gain
I'm currently at about -40% CPU and -10% memory
And haven't done the in-kernel dedup yet ...

pkg/model/record_test.go
```diff
@@ -1,107 +0,0 @@
-package model
```
jotak, line 1: hey @msherif1234 , I'm removing this test because I think it is of quite low value: it only tests `ReadFrom`, which is really a direct call to `binary.Read`. Meaning all it does is testing that the `encoding/binary` standard lib is right. This, at a price of sometimes painful maintenance. Does it worth it?

bpf/flows.c
```diff
@@ -107,28 +108,13 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
         aggregate_flow->dns_record.flags = pkt.dns_flags;
         aggregate_flow->dns_record.latency = pkt.dns_latency;
         aggregate_flow->dns_record.errno = dns_errno;
-        long ret = bpf_map_update_elem(&aggregated_flows, &id, aggregate_flow, BPF_ANY);
```
jotak, line 110: this part (removing unnecessary bpf_map_update_elem) is left incomplete on purpose, I'll wait that @msherif1234 's PR is merged then rebase this one

tohojo:
Looked through the BPF code, and it basically LGTM :)

bpf/flows.c
```diff
@@ -93,6 +93,7 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
     // a spinlocked alternative version and use it selectively https://lwn.net/Articles/779120/
     flow_metrics *aggregate_flow = (flow_metrics *)bpf_map_lookup_elem(&aggregated_flows, &id);
     if (aggregate_flow != NULL) {
+        bpf_spin_lock(&aggregate_flow->lock);
```
msherif1234, line 96: we probably need to check the return code for spin lock and unlock and log an error

bpf/flows.c
```diff
@@ -80,7 +80,7 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
     id.direction = direction;
 
     // check if this packet need to be filtered if filtering feature is enabled
-    bool skip = check_and_do_flow_filtering(&id, pkt.flags, 0);
+    bool skip = check_and_do_flow_filtering(&pkt, 0);
```
msherif1234, line 83: I don't understand the need to replace id with pkt struct that generated lots of unnecessary changes and I also planned to even delete the pkt info struct at some point ?

bpf/flows.c
```diff
@@ -137,12 +123,13 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
...
             .dns_record.errno = dns_errno,
-            .flow_rtt = rtt,
         };
+        __builtin_memcpy(new_flow.dst_mac, pkt.dst_mac, ETH_ALEN);
+        __builtin_memcpy(new_flow.src_mac, pkt.src_mac, ETH_ALEN);
 
         // even if we know that the entry is new, another CPU might be concurrently inserting a flow
         // so we need to specify BPF_ANY
-        long ret = bpf_map_update_elem(&aggregated_flows, &id, &new_flow, BPF_ANY);
+        long ret = bpf_map_update_elem(&aggregated_flows, &id, &new_flow, BPF_NOEXIST);
```
msherif1234, line 132: we will probably need to spin lock/unlock here as well

bpf/flows.c
msherif1234: this the min value of ythe hook isn't fire IIRC it was a define in the kernel and I believe we doc that as the default min RTT, when I recall more I will update here

bpf/types.h
```diff
@@ -89,37 +89,30 @@ typedef struct flow_metrics_t {
     // https://chromium.googlesource.com/chromiumos/docs/+/master/constants/errnos.md
     u8 errno;
     u8 dscp;
+    struct dns_record_t {
+        u16 id;
+        u16 flags;
+        u64 latency;
+        u8 errno;
+    } dns_record;
```
msherif1234, line 97: why we keep dns record here and not move to additional metrics ?

bpf/types.h
```diff
@@ -238,7 +218,20 @@ struct filter_value_t {
     u8 filter_drops;
     u8 ip[IP_MAX_LEN];
 } __attribute__((packed));
-// Force emitting struct filter_value_t into the ELF.
-const struct filter_value_t *unused9 __attribute__((unused));
+
+// Force emitting enums/structs into the ELF
+const enum tcp_flags_t *unused0 __attribute__((unused));
```
msherif1234, line 223: I intentionally added that line after each structure definition as reminder so bpf2go can generate eqv go struct for userspace

bpf/flows.c
jotak: this is due to removing stuff from flow_id, and moving them to flow_metrics, like MAC addresses. As a result, id isn't sufficient to pass it to all these functions in `utils`. So instead of adding tons of arguments to the signatures I just use the `pkt` that already has everything needed

Makefile
```diff
@@ -106,6 +106,7 @@ prereqs: ## Check if prerequisites are met, and install missing dependencies
 fmt: ## Run go fmt against code.
 	@echo "### Formatting code"
 	go fmt ./...
+	find ./bpf -type f -not -path "./bpf/headers/*" -name "*.[ch]" | xargs clang-format -i --Werror
```
msherif1234, line 109: I have this rule as part of the lint target we can make it part of the fmt but in that case we need to remove it from the lint so we won't run it twice when build ?

bpf/flows.c
jotak: `pkt` is convenient to not have extra long signatures

bpf/types.h
jotak: yeah I mentioned that in the TODO here https://github.com/netobserv/netobserv-ebpf-agent/pull/470
I didn't do it at first because it's modified from the TC hook so it isn't required to move it ... but still it would probably be better to do it.

bpf/types.h
jotak: They're just moved at the end of the file. I can rollback that, I don't really care, but I was finding it easier to choose the next "unusedX" number when having all of them in sight at once

bpf/flows.c
jotak: the lock is on the value object, not the key, and it's a new object in that case, so I don't think it makes sense to lock

bpf/flows.c
msherif1234: u just removed l2 and interface right ? or am I missing something else ? why would flow filter or util core apis need to change ?

bpf/types.h
msherif1234: not grouping them to gether make it easy to forget but at the cost of looking for the next unused manually

bpf/flows.c
tohojo: That's what the `BPF_NOEXIST` flag does: If two CPUs call this concurrently, one of them will fail with EEXIST, and then fall back to the lookup/update path (which does take the lock)

msherif1234: with eexist lookup and update we need to take the lock as well what about `BPF_F_LOCK` can I set that flag to takecare of the lock/unlock for me ?

tohojo: `BPF_F_LOCK` is for when you want to replace the whole contents of the map value; but that's not what you're doing, you are increasing a couple of counters. So you just re-use exactly the same update code as for when the map lookup succeeds in the first place :)

bpf/flows.c
jotak: and `eth_protocol` and `direction`

bpf/types.h
jotak: ok i'll rollback that

bpf/flows.c
msherif1234: remember datapath has to be efficient adding additional dereferrence isn't efficient IMO that stuff not used and I don't think will impact the apis to change the signature much

jotak: I'm reaching the 5 args limit of ebpf calls if I want to remove them from `pkt`

bpf/flows.c
jotak: actually, there's no error code if I'm correct. The doc for bpf_spin_lock says
```
 * Returns
 * 	0
```
In linux helper.c it returns void: https://elixir.bootlin.com/linux/v6.12.1/source/kernel/bpf/helpers.c#L283

bpf/flows.c
jotak: fixed via https://github.com/netobserv/netobserv-ebpf-agent/pull/469/commits/8e69cfd4a3703561ac26b8218d4079aa11e25edc 
that allows also to get rid of a few unnecessary MAC copying

bpf/types.h
jotak: addressed via https://github.com/netobserv/netobserv-ebpf-agent/pull/469/commits/01950d5b862e9ce494de90caf87a8846baa4b89d

bpf/types.h
jotak: addressed via https://github.com/netobserv/netobserv-ebpf-agent/pull/469/commits/01950d5b862e9ce494de90caf87a8846baa4b89d

Makefile
jotak: I think that's for two different things: the format rule allows to auto-format as part of the _local_ build process on your machine, whereas the lint rule is still useful _on the CI_ to make sure we didn't commit & push unformatted code.

jotak:
I appear to have broken DNS somehow ...

bpf/flows.c
msherif1234: doc said it was returning something https://docs.ebpf.io/linux/helper-function/bpf_spin_lock/ maybe doc is wrong

bpf/flows.c
```diff
@@ -51,13 +51,27 @@
  */
 #include "network_events_monitoring.h"
 
+static inline void update_dns(additional_metrics *extra_metrics, pkt_info *pkt, int dns_errno) {
+    if (pkt->dns_id != 0) {
```
msherif1234, line 55: u can check dns_err here and if its `0` then add DNS info else add dns_error

bpf/flows.c
jotak: it does, but apparently it's always 0, not sure why : https://github.com/netobserv/netobserv-ebpf-agent/blob/main/bpf/headers/bpf_helper_defs.h#L2359-L2360

bpf/flows.c
jotak: there can be an error and still have DNS data: when the dns request wasn't found, we can't generate latency, but we still provide the other data, and return ENOENT.
So it shouldn't be a if/else here, if we want to still provide what we have

jotak:
Perf-scale test results: [comparison](https://docs.google.com/spreadsheets/d/1sqcbIOtNTylX74gsgQo9Q75krAiZJiCxmZ_oZdEL2hk/edit?gid=557066635#gid=557066635), [full data sheet](https://docs.google.com/spreadsheets/d/1ZpAnepjVuCRfoXqhn-TLjS-sFqYk8YuT_z_8WJeSaWM/edit?gid=1538180928#gid=1538180928)
Comparison is done with https://github.com/netobserv/netobserv-ebpf-agent/pull/466

=> -43% agent CPU ; other metrics are roughly stable (~+2 to 5% in everything, including workload throughput, number of flows, memory used, etc.)

msherif1234:
> Perf-scale test results: [comparison](https://docs.google.com/spreadsheets/d/1sqcbIOtNTylX74gsgQo9Q75krAiZJiCxmZ_oZdEL2hk/edit?gid=557066635#gid=557066635), [full data sheet](https://docs.google.com/spreadsheets/d/1ZpAnepjVuCRfoXqhn-TLjS-sFqYk8YuT_z_8WJeSaWM/edit?gid=1538180928#gid=1538180928) Comparison is done with #466
> 
> => -43% agent CPU ; other metrics are roughly stable (~+2 to 5% in everything, including workload throughput, number of flows, memory used, etc.)

its very strange that memory shows tiny increase moving from perCPU to global for the same number of flows I would expected to see ~ 1/ncpu memory drop is the cpu drop mostly in lookupAndDelete() or somewhere else ?

tohojo:
"Mohamed S. Mahmoud" ***@***.***> writes:

>> => -43% agent CPU ; other metrics are roughly stable (~+2 to 5% in everything, including workload throughput, number of flows, memory used, etc.)
>
> its very strange that memory shows tiny increase moving from perCPU to global for the same number of flows I would expected to see ~ 1/ncpu memory drop is the cpu drop mostly in lookupAndDelete() or somewhere else ?

Are you actually measuring the memory and CPU usage of the BPF programs
and maps, or is this just numbers for the userspace side? If it's the
latter I guess it makes sense that the memory usage doesn't change, and
the drop in CPU would be because the time spent de-duplicating things
goes away?

msherif1234:
> "Mohamed S. Mahmoud" ***@***.***> writes:
> > => -43% agent CPU ; other metrics are roughly stable (~+2 to 5% in everything, including workload throughput, number of flows, memory used, etc.) its very strange that memory shows tiny increase moving from perCPU to global for the same number of flows I would expected to see ~ 1/ncpu memory drop is the cpu drop mostly in lookupAndDelete() or somewhere else ?
> Are you actually measuring the memory and CPU usage of the BPF programs and maps, or is this just numbers for the userspace side? If it's the latter I guess it makes sense that the memory usage doesn't change, and the drop in CPU would be because the time spent de-duplicating things goes away?

we still doing dedup in this PR but we no longer aggregate we probably need to compare memory using bpftool to also check kernel usage

jotak:
Yes exactly we don't track today the BPF memory :-/ We definitely should

tohojo:
Joel Takvorian ***@***.***> writes:

> Yes exactly we don't track today the BPF memory :-/ We definitely
> should

The BPF map memory usage is accounted to memory cgroup of the process
that creates the map, so one way to do this would be to just record
memory usage at the cgroup level; then you won't have to do anything
special for BPF :)

bpf/network_events_monitoring.h
```diff
@@ -35,10 +35,9 @@ static inline int lookup_and_update_existing_flow_network_events(flow_id *id, u8
 
     bpf_probe_read(cookie, md_len, user_cookie);
 
-    flow_metrics *aggregate_flow = bpf_map_lookup_elem(&aggregated_flows, id);
+    additional_metrics *aggregate_flow = bpf_map_lookup_elem(&additional_flow_metrics, id);
```
msherif1234, line 38: nit: var name is a bit misleading here maybe flow_feature or something I assume on all other features too

bpf/network_events_monitoring.h
```diff
@@ -113,19 +107,12 @@ static inline int trace_network_events(struct sk_buff *skb, struct rh_psample_me
...
-    id.direction = INGRESS;
-    flow_metrics new_flow = {
-        .packets = 1,
-        .bytes = len,
-        .start_mono_time_ts = current_time,
-        .end_mono_time_ts = current_time,
-        .flags = flags,
+    additional_metrics new_flow = {
         .network_events_idx = 0,
     };
```
msherif1234, line 112: there are instances where pkt hit the hook but not the tcx in that case u will see flows with o bytes and pkts did u check for that ?

bpf/flows.c
```diff
@@ -171,6 +174,48 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
             }
         }
     }
+
+    // Update additional metrics (per-CPU map)
+    if (pkt.dns_id != 0 || dns_errno != 0) {
+        // hack on id will be removed with dedup-in-kernel work
+        id.direction = 0;
+        id.if_index = 0;
```
msherif1234, line 182: will the feature work with this hack ?

bpf/types.h
```diff
@@ -138,24 +149,21 @@ typedef struct flow_id_t {
     u8 icmp_code;
     // OS interface index
     u32 if_index;
-} __attribute__((packed)) flow_id;
+} flow_id;
```
msherif1234, line 152: without packed the generated golang structure will have padding and will be hard to predict the size of the struct which might fail unit test

pkg/flow/tracer_map.go
```diff
@@ -102,14 +102,13 @@ func (m *MapTracer) evictFlows(ctx context.Context, forceGC bool, forwardFlows c
 	flows := m.mapFetcher.LookupAndDeleteMap(m.metrics)
 	elapsed := time.Since(currentTime)
 	for flowKey, flowMetrics := range flows {
-		aggregatedMetrics := m.aggregate(flowMetrics)
 		// we ignore metrics that haven't been aggregated (e.g. all the mapped values are ignored)
```
msherif1234, line 105: nit: irrelevant comment

pkg/model/flow_content.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	for _, md := range other.NetworkEvents {
+		if !AllZerosMetaData(md) && !networkEventsMDExist(p.AdditionalMetrics.NetworkEvents, md) {
+			copy(p.AdditionalMetrics.NetworkEvents[p.AdditionalMetrics.NetworkEventsIdx][:], md[:])
+			p.AdditionalMetrics.NetworkEventsIdx = (p.AdditionalMetrics.NetworkEventsIdx + 1) % MaxNetworkEvents
+		}
+	}
+}
+
+func allZero(s [6]uint8) bool {
```
msherif1234, line 95: lets call allZerosMac()

pkg/model/flow_content.go
```diff
@@ -0,0 +1,102 @@
+package model
+
+import "github.com/netobserv/netobserv-ebpf-agent/pkg/ebpf"
+
+type BpfFlowContent struct {
+	ebpf.BpfFlowMetrics
```
msherif1234, line 6: can u use ptr here as well ?

pkg/model/record_test.go
msherif1234: what is the reason u removing it because u removed packed so the structure no longer aligns ?

msherif1234: the advantage of this make sure the auto generated structure are sane too

pkg/tracer/tracer.go
```diff
@@ -715,39 +722,57 @@ func (m *FlowFetcher) ReadRingBuf() (ringbuf.Record, error) {
...
 			continue
 		}
-		flows[id] = metrics
+		flowPayload := model.BpfFlowContent{}
+		flowPayload.AccumulateBase(&baseMetrics)
+
+		// Fetch additional metrics; ids are without direction and interface
+		shorterID := id
+		shorterID.Direction = 0
+		shorterID.IfIndex = 0
```
msherif1234, line 764: is that related to hack in the bpf code where they aren't set ?

bpf/network_events_monitoring.h
jotak: sure - in other places I called that `extra_metrics` so will use the same for consistency

bpf/network_events_monitoring.h
jotak: what sort of instances? My problem with that is it would duplicate counts, if this hook is called before the tc - or if the TC flow was just flushed - hence leading to over-estimate. I would rather keep strict responsibility boundaries : TC for counters, other hooks for additional features, and not mix up roles.

bpf/flows.c
jotak: sure, it's precisely here to make it work :-)
On user-space we assume all additional map entries are without dir/interface. This is temporary, my other PR is removing that, since eventually we want to entirely get rid of interface/direction in flow key.

bpf/types.h
jotak: Removing packed was one of the things that really improved performances. And also the spin lock can't be in a packed struct (or at least, linter complains, I guess for good reasons?). So having a unit-test harder to maintain is IMO a lesser concern? (btw I want to remove that test but I saw you're commenting on that)

pkg/flow/tracer_map.go
jotak: in fact, I don't quite understand what follows.. Do you?
```
		// we ignore metrics that haven't been aggregated (e.g. all the mapped values are ignored)
		if flowMetrics.EndMonoTimeTs == 0 {
			continue
		}
```
Are there actual cases where end time can be 0? Or was there, before this refactoring? I don't think so, but preferred to leave it as is ...

pkg/tracer/tracer.go
jotak: correct

pkg/model/record_test.go
jotak: I still don't get, what sort of issue could it catch, do you have an example?
Anyway, I'm fine reintroducing it. It was not only for the padding change, it's any change in the bpf types that make this test painful to maintain.

jotak: Here we go: https://github.com/netobserv/netobserv-ebpf-agent/pull/469/commits/119061c05dbafe2dc07e45bd34419d3e58be23c2
But I'd still appreciate if you can give me a concrete example. It doesn't really test anything about the types.h / go struct mapping, it only test that a go struct is what it claims to be ... that doesn't make really sense to me.

bpf/network_events_monitoring.h
jotak: done

pkg/model/flow_content.go
jotak: done

pkg/model/flow_content.go
jotak: done

pkg/model/record.go
```diff
@@ -23,7 +23,7 @@ const MacLen = 6
 const (
 	IPv6Type                 = 0x86DD
 	networkEventsMaxEventsMD = 8
```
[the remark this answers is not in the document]
jotak: done

bpf/flows.c
```diff
@@ -63,10 +63,18 @@ static inline void update_existing_flow(flow_metrics *aggregate_flow, pkt_info *
     }
     aggregate_flow->flags |= pkt->flags;
     aggregate_flow->dscp = pkt->dscp;
-    aggregate_flow->dns_record.id = pkt->dns_id;
-    aggregate_flow->dns_record.flags = pkt->dns_flags;
-    aggregate_flow->dns_record.latency = pkt->dns_latency;
-    aggregate_flow->dns_record.errno = dns_errno;
+    bpf_spin_unlock(&aggregate_flow->lock);
```
msherif1234, line 66: we no longer needs https://github.com/netobserv/netobserv-ebpf-agent/blob/119061c05dbafe2dc07e45bd34419d3e58be23c2/bpf/flows.c#L59
as that was specific to perCPU map

bpf/types.h
msherif1234: do u recall what was the linter error ? plus not very clear on how removing pack would help performance that much tbh, I think generating predicitable userspace struct is important but let see the linter error 1st

pkg/ebpf/bpf_arm64_bpfel.go
```diff
@@ -75,35 +87,35 @@ type BpfFilterValueT struct {
...
-	DstMac            [6]uint8
 	SrcIp             [16]uint8
 	DstIp             [16]uint8
+	_                 [1]byte
 	SrcPort           uint16
 	DstPort           uint16
 	TransportProtocol uint8
 	IcmpType          uint8
 	IcmpCode          uint8
+	_                 [3]byte
```
msherif1234, line 99: this example of what removing packed will cause BTW we could avoid so gaps by moving the fields around if u move direction to the bottom might help this struct ?

pkg/exporter/ipfix.go
```diff
@@ -294,31 +294,31 @@ func setIEValue(record *model.Record, ieValPtr *entities.InfoElementWithValue) {
...
+		ieVal.SetUnsigned16Value(record.ID.SrcPort)
 	case "destinationTransportPort":
-		ieVal.SetUnsigned16Value(record.Id.DstPort)
+		ieVal.SetUnsigned16Value(record.ID.DstPort)
 	case "icmpTypeIPv4", "icmpTypeIPv6":
-		ieVal.SetUnsigned8Value(record.Id.IcmpType)
+		ieVal.SetUnsigned8Value(record.ID.IcmpType)
 	case "icmpCodeIPv4", "icmpCodeIPv6":
-		ieVal.SetUnsigned8Value(record.Id.IcmpCode)
+		ieVal.SetUnsigned8Value(record.ID.IcmpCode)
```
msherif1234, line 321: this change from `Id` to `ID` generated tons of unnecessary changes here in many other places

msherif1234:
few nits but I am more less done with review, will be good to run features testing with this PR pre merge
/lgtm

msherif1234:
pls update the arch doc reference to perCPu hashmap and make sure to run https://github.com/netobserv/netobserv-ebpf-agent/pull/172#issuecomment-1698837448 to make sure we have no concurrency issues again

bpf/flows.c
jotak: correct, good catch

pkg/ebpf/bpf_arm64_bpfel.go
jotak: so in this PR I haven't care of these little optimizations because in the next PR I still move stuff around and [there](https://github.com/netobserv/netobserv-ebpf-agent/pull/470/commits/f0d743c47c838c5666b6d9cbd5fc37069dbd7998#diff-dbaedeecfc6f45e9f0f7521e5146144ca41b968d50b0acad82393f70d2b2e20f) I'm caring about these optimizations

bpf/types.h
jotak: (I don't have in mind what the error was, but definitely was something wrong with lock+packed)

bpf/network_events_monitoring.h
msherif1234: then u need to make sure u have no rows in the UI with no counters as that is misleading

pkg/exporter/ipfix.go
jotak: yeah but it's the naming convention & makes golint happier ...

bpf/flows.c
jotak: fixed

pkg/flow/tracer_map.go
jotak: fixed

jotak:
btw added a diagram for kernel space https://github.com/jotak/netobserv-agent/blob/global-map/docs/architecture.md

bpf/types.h
msherif1234: I found this https://github.com/cilium/ebpf/blob/f0eec7efba9dd069f4d937d3096e276c7911a1c0/memory.go#L26 that is very likely explains the userspace cpu improvements

pkg/exporter/ipfix.go
msherif1234: might make linter happier but screw up all inflightes PRs :(

msherif1234:
> btw added a diagram for kernel space https://github.com/jotak/netobserv-agent/blob/global-map/docs/architecture.md

can u share the tool used to draw as we add new hooks on a regular basis so that diag will be quickly become out of date

jotak:
@msherif1234 you don't necessarily need a tool you can directly edit script:

```
flowchart TD
    A[TC/X Hooks] -->|Accumulate packet data| M1(Global map: aggregated_flows)
    D{If DNS} -->|Req: store req info| MD(Global map: dns_flows)
    D -->|Resp: compute latency| MD
    A -->D
    D -->|Store DNS info| M2(PerCPU map: additional_flow_metrics)
    B[Drops Hook: kfree_skb] -->|Accumulate drop data| M2(PerCPU map: additional_flow_metrics)
    C[RTT Hook: tcp_rcv_established] -->|Extract & store sRTT| M2(PerCPU map: additional_flow_metrics)
    E[Events Hook: psample_sample_packet] -->|Accumulate net events| M2(PerCPU map: additional_flow_metrics)
    A -->F{If busy map / error}
    F -->|Single-packet flow| RB(RingBuffer)
    M1 --> |Polling|U[User space]
    M2 --> |Polling|U
    RB --> |Push|U
```

So you can easily add a new hook by adding something like:

```
F[My new hook] --> |any interaction written here| M2
```
(M2 stands for the additional metrics map)

Or if you prefer there's also online tools such as this: https://mermaid.live/

jotak:
I took some measurements on the ebpf side, written down there: https://docs.google.com/spreadsheets/d/1gdwX4ZLjKicaJzANVq-r-lAW6uZxefqjBsIw3Orx4bo/edit?gid=0#gid=0
That's on a single (worker) node, running my usual stress tests. I looked at the sum of each data structure size, then removed a constant 16MB that's taken by ring buffer to make the gains on maps more apparent.
For CPU I used [bpftop](https://github.com/Netflix/bpftop), with very approximative manual measurement (don't rely too much on my CPU measurements!)

**Between -28% and -54% memory** used depending on scenarios. And **between -4% and -36% CPU**.

The [next PR](https://github.com/netobserv/netobserv-ebpf-agent/pull/470) improves that a little bit further with **between -42% and -61% memory**, and **around -20% CPU**.

jotak:
thanks @msherif1234  @tohojo @Amoghrd !
/approve

PR #470, opened 2024-11-29
NETOBSERV-1996: in-kernel de-duplication

jotak:
## Description
    
- Remove interface from the flow key; instead, use it as flow value. The
      first interface+dir seen for a given flow is the one taken into account
    for counters. Other interfaces+dirs are stored in a separate map for this
    flow. This algorithm is more or less the deduper algo that we had in
    userspace.
- Remove user-space deduper
- Adapt user-space model for the new interfaces+directions array
      provided directly from ebpf structs
- Remove "decorator" (which was doing the interface naming). This is
      just for simplification. This enrichment is now done in a more
    straightforward way, when creating the Record objects

(this PR is based on https://github.com/netobserv/netobserv-ebpf-agent/pull/466 and https://github.com/netobserv/netobserv-ebpf-agent/pull/469)

TODO:

- [x] Move DNS info out of the main map. New map of its own? Maybe split the maps even more (1 per feature)
- [ ] Think more about map sizing (should secondary maps have same max size as main map?)
- [ ] Remove RB? cf discussion below https://github.com/netobserv/netobserv-ebpf-agent/pull/470#discussion_r1863619427

## Dependencies

n/a

## Checklist

* [x] Is this PR backed with a JIRA ticket? If so, make sure it is written as a title prefix _(in general, PRs affecting the NetObserv/Network Observability product should be backed with a JIRA ticket - especially if they bring user facing changes)._
  * [x] Regression tests only (e.g. refactoring with no user-facing change).


bpf/flows.c
```diff
@@ -51,6 +51,40 @@
...
+    if (value->nb_observed_intf < MAX_OBSERVED_INTERFACES) {
+        for (u8 i = 0; i < value->nb_observed_intf; i++) {
+            if (value->observed_intf[i].if_index == if_index &&
+                value->observed_intf[i].direction == direction) {
+                return;
+            }
+        }
+        value->observed_intf[value->nb_observed_intf].if_index = if_index;
+        value->observed_intf[value->nb_observed_intf].direction = direction;
+        value->nb_observed_intf++;
```
tohojo, line 84: The update should be under the spinlock, or you can have two concurrent updates writing different values to the same slot.

You'll have to re-do the check while holding the lock to make sure things are really consistent.

jotak: this map is the multi-cpu one, so no need for spinlock, no?

jotak: well, I realize I'm shadowing the `aggregate_flow` variable with another one coming from a different map ... which can cause confusion, I'll change that

bpf/flows.c
```diff
@@ -137,34 +156,46 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
...
+                if (!record) {
+                    if (trace_messages) {
+                        bpf_printk("couldn't reserve space in the ringbuf. Dropping flow");
+                    }
+                    return TC_ACT_OK;
                 }
-                return TC_ACT_OK;
+                record->id = id;
+                record->metrics = new_flow;
+                bpf_ringbuf_submit(record, 0);
```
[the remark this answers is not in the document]
jotak: That's a good point, this ringbuf is clearly double-edged. But before doing so I'd like to re-investigate actual reasons for switching to ringbuf, ie. if it's purely a matter of map capacity then I agree users can simply adapt the map config; but if there are other reasons, without map being at full capacity, perhaps it still makes sense?

bpf/flows.c
tohojo: Ah, yes, the differently-typed variables with the same name threw me off. No, in that case, you don't need to lock :)

bpf/flows.c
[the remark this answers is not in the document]
jotak: cc @msherif1234 wdyt about that?
In my stress tests, the rb:hm ratio is typically at most 0.0001 ie. one flow emitted by RB for 10.000 flows emitted by hashmap
Looking at the perf-scale tests, this is quite similar, ingress-perf shows 0.0005 and cluster-density 0.00002
So, dropping the RB would like be unnoticed anyway....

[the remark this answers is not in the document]
tohojo: Sure, if there's another reason why you may fall back to the ringbuffer, it could make sense to keep it. But with the dedup happening in the kernel, there really shouldn't be? E2BIG means the map is full, and ENOMEM means the system is under memory pressure. The one other error your comment mentions (EBUSY) can only happen when the map is concurrently accessed multiple times *on the same CPU*; which shouldn't happen from the TC hook.

So *if* you do see errors that trigger the fallback path, I would be very interested in understanding exactly why that happens :)

jotak:
[Perf results](https://docs.google.com/spreadsheets/d/1qakBaK1dk_rERO30k1cSR4W-Nn0SXW4A3lqQ1sZC4rE/edit?gid=994867402#gid=994867402) ~ -60% CPU and -10% memory

bpf/flows.c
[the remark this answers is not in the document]
tohojo: Huh. And that is with this PR applied?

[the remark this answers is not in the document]
jotak: yes, between 13:20 and 14:10 this PR was applied (after that it's without this PR)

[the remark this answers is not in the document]
tohojo: Hmm, okay, that's really odd. Just to double check, which kernel version is this running on?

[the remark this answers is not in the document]
jotak: I didn't check but it was an ocp 4.17 so rhel-9.4 if I'm correct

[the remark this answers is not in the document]
jotak: Could it be because of not using `BPF_F_LOCK` while doing the lookup&delete on user space side? I'm just reading about it, I haven't updated the user-space lookup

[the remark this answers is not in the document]
jotak: sounds like it can be any interface and direction: 
image
So I have some ens5, br-ex, genev_sys_6081, 654cc34c0d2792b@if2 (ie. a pod interface)

jotak:
Perf-scale test results: [comparison](https://docs.google.com/spreadsheets/d/1sYjOBES0T2Z-T9o2JJKmACjqRSfi_1XXW4mExijbR3Y/edit?gid=461818885#gid=461818885), [full data sheet](https://docs.google.com/spreadsheets/d/196muMg_jbt6AfS368Hyk_y99UPEoRBA7yYdEp3PQlqE/edit?gid=277656009#gid=277656009)
Comparison is done with https://github.com/netobserv/netobserv-ebpf-agent/pull/469
And I added a comparison with main / https://github.com/netobserv/netobserv-ebpf-agent/pull/466

=> -33% agent CPU, -14% agent memory (user-space); and also +10% workload throughput. Compared to main, agent shows -62% CPU and -12% memory, +13% workload throughput. Altogether, overall netobserv user-space memory is slightly improved, -5%

jotak:
/hold
something wrong happened during a rebase; it produces much higher output than it should. That's what the e2e tests are saying and it's also something I saw manually.
Earlier version (before rebases) was working fine

bpf/types.h
```diff
@@ -130,23 +133,23 @@ typedef struct additional_metrics_t {
         u16 zone_id;
         u8 icmp_id;
     } translated_flow;
+    struct observed_intf_t {
+        u8 direction;
+        u32 if_index;
+    } __attribute__((packed)) observed_intf[MAX_OBSERVED_INTERFACES];
```
msherif1234, line 139: `packed` is bad for userspace cpu

bpf/flows.c
```diff
@@ -125,10 +136,25 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
...
+        } else {
+            // Only add info that we've seen this interface
+            additional_metrics *extra_metrics =
+                (additional_metrics *)bpf_map_lookup_elem(&additional_flow_metrics, &id);
+            if (extra_metrics != NULL) {
+                add_observed_intf(extra_metrics, skb->ifindex, direction);
+            } else {
+                additional_metrics new_metrics = {};
+                add_observed_intf(&new_metrics, skb->ifindex, direction);
+                bpf_map_update_elem(&additional_flow_metrics, &id, &new_metrics, BPF_NOEXIST);
```
msherif1234, line 150: check and handle error cases and EEXISTS error

bpf/flows.c
```diff
@@ -78,6 +78,20 @@ static inline void update_dns(additional_metrics *extra_metrics, pkt_info *pkt,
     }
 }
 
+static inline void add_observed_intf(additional_metrics *value, u32 if_index, u8 direction) {
+    if (value->nb_observed_intf < MAX_OBSERVED_INTERFACES) {
+        for (u8 i = 0; i < value->nb_observed_intf; i++) {
+            if (value->observed_intf[i].if_index == if_index &&
```
msherif1234, line 84: we should check `if_index` make sure its not 0

jotak: hmm do you think I should also ignore setting `if_index_first_seen` to `skb->ifindex` if it's 0 ?
(in 
```
        flow_metrics new_flow = {
            .if_index_first_seen = skb->ifindex,
            .direction_first_seen = direction,
            .packets = 1,
            .bytes = len,
// ...
```
)

jotak: (idk in what cases it would be 0 and if it's something to handle in a special way)

jotak:
@msherif1234 thanks for the review, comments addressed

bpf/flows.c
jotak: added `else if (skb->ifindex != 0)` earlier in the code path

bpf/types.h
jotak: good catch, done

bpf/flows.c
jotak: done

bpf/flows.c
msherif1234: there are certian hooks that don't populate if_index u can see examples of that in the current code where we skip populating flow if

bpf/flows.c
```diff
@@ -78,6 +78,20 @@ static inline void update_dns(additional_metrics *extra_metrics, pkt_info *pkt,
...
+                value->observed_intf[i].direction == direction) {
+                return;
+            }
+        }
+        value->observed_intf[value->nb_observed_intf].if_index = if_index;
+        value->observed_intf[value->nb_observed_intf].direction = direction;
+        value->nb_observed_intf++;
+    }
```
msherif1234, line 92: we need an else here and log a msg and probably updated a counter I assume the value of 4 is based on what we see today but depends on how often we come to else we might need to increase it.

what happens when we hit the above condition is that a bad flow that we should drop or u can return error code in this function and do the print and counter update at the caller but we can't just be silent when this condition happens

pkg/model/flow_content.go
```diff
@@ -98,6 +94,26 @@ func (p *BpfFlowContent) AccumulateAdditional(other *ebpf.BpfAdditionalMetrics)
 	if !AllZeroIP(IP(other.TranslatedFlow.Saddr)) && !AllZeroIP(IP(other.TranslatedFlow.Daddr)) {
 		p.AdditionalMetrics.TranslatedFlow = other.TranslatedFlow
 	}
+	// Accumulate interfaces + directions
+	accumulateInterfaces(&p.AdditionalMetrics.NbObservedIntf, &p.AdditionalMetrics.ObservedIntf, other.NbObservedIntf, other.ObservedIntf)
+}
+
+func accumulateInterfaces(dstSize *uint8, dstIntf *[MaxObservedInterfaces]ebpf.BpfObservedIntfT, srcSize uint8, srcIntf [MaxObservedInterfaces]ebpf.BpfObservedIntfT) {
```
msherif1234, line 101: can u add UT for this make sure it handles all variances

jotak: it's actually already covered but in another package, I'll move the test to the appropriate package

bpf/flows.c
jotak: right, +1

jotak: I think there's the same thing with MAX_NETWORK_EVENTS that we may silently ignore

msherif1234: network events I return error so that should be fine

jotak: Oh I didn't notice the rolling increase [here](https://github.com/netobserv/netobserv-ebpf-agent/blob/main/bpf/network_events_monitoring.h#L45)
So when capacity is reached, it erases the oldest entry, but it do so silently, or am I missing something?

msherif1234: yeah that was by design newer events will overwrite older ones but as of now we can't get more than 3 events max per flow so we should be fine for now

jotak: done here https://github.com/netobserv/netobserv-ebpf-agent/pull/470/commits/f2a7a94fe1b7bbf9791ebbe964ca5e13e7d7caea

pkg/model/flow_content.go
jotak: test moved: https://github.com/netobserv/netobserv-ebpf-agent/pull/470/commits/52341d440775e6b31aa00d7bc751b7c44c360490

bpf/flows.c
jotak: ok but here it's the tc hook ; maybe that's something not relevant now that other hooks use a different map?  anyway it won't hurt to keep these checks, I guess...

pkg/tracer/tracer_legacy.go
```diff
@@ -14,24 +14,20 @@ func (m *FlowFetcher) legacyLookupAndDeleteMap(met *metrics.Metrics) map[ebpf.Bp
...
 	// Deleting while iterating is really bad for performance (like, really!) as it causes seeing multiple times the same key
 	// This is solved in >=4.20 kernels with LookupAndDelete
-	for iterator.Next(&id, &metrics) {
+	for iterator.Next(&id, &baseMetrics) {
 		count++
 		if err := flowMap.Delete(id); err != nil {
 			log.WithError(err).WithField("flowId", id).Warnf("couldn't delete flow entry")
 			met.Errors.WithErrorName("flow-fetcher-legacy", "CannotDeleteFlows").Inc()
 		}
 		// We observed that eBFP PerCPU map might insert multiple times the same key in the map
```
msherif1234, line 28: nit: baseMetrics is no longer perCPU so this outdated comment

msherif1234:
another round of doc update is needed after deduper block is removed

msherif1234:
I am done reviewing this again good round of testing is needed to minimize regressions
/lgtm

pkg/tracer/tracer_legacy.go
jotak: done

jotak:
Thanks @msherif1234 @Amoghrd 
I also updated the doc diagram

PR #472, opened 2024-12-04
WIP: kernel map iterator to evict stale flows and avoid hmap churns

msherif1234:
## Dependencies

n/a

## Checklist


bpf/flows.c
```diff
@@ -61,6 +66,10 @@ static inline void update_existing_flow(flow_metrics *aggregate_flow, pkt_info *
     if (aggregate_flow->start_mono_time_ts == 0) {
         aggregate_flow->start_mono_time_ts = pkt->current_ts;
     }
+
+    if (aggregate_flow->last_seen_timestamp == 0) {
```
tohojo, line 70: This should be `if (pkt->currents_ts != 0)` - i.e., always update the timestamp with the latest value.

pkg/tracer/tracer.go
```diff
@@ -756,6 +769,65 @@ func (m *FlowFetcher) LookupAndDeleteMap(met *metrics.Metrics) map[ebpf.BpfFlowI
...
+	}
+
+	// trigger iterator kernel check to look for stale entries
+	f, err := m.flowMapLink.Open()
+	if err != nil {
+		log.WithError(err).WithField("flowId", id).Warnf("couldn't open flows map iterator")
+		m.lookupAndDeleteIterateSupported = false
+		return m.lookupAndDeleteMapUserSpace(met)
+	}
+	f.Close()
```
tohojo, line 823: It's not enough to open and close the link, you also need to read from it. See: 

https://github.com/xdp-project/bpf-examples/blob/master/pping/pping.c#L696

tohojo (commented):
A couple of nits on the usage of the iterator. But as a wider point, I'm not sure if this approach will give you the right values. Flows that don't expire will have their full counters read on every poll, so you'll get aggregate counter values instead of the values since last poll. So the aggregation logic will have to take this into account, and if you end up calculating the delta in userspace afterwards, I think that will eat any gains there may be from this...

pkg/tracer/tracer.go
msherif1234: ok I know I was missing something with iterator userspace trigger

msherif1234:
image

msherif1234:
this approach doesn't seem to yield in good resources usage vs doing it all in userspace
image

msherif1234:
closing this PR as it turned out it won't really help much the userspace resources

PR #483, opened 2024-12-13
Add enrichment stats

jotak:
New metric: `netobserv_agent_flows_enrichment_total`

The goal is to be able to track regressions in terms of quantity of flows being enriched.

Example of query:
`sum(rate(netobserv_agent_flows_enrichment_total[1m])) by (hasDNS, hasRTT, hasDrops, hasNetEvents)`

image

Or, to have the proportion of flows with RTT for instance:
`100 * sum(rate(netobserv_agent_flows_enrichment_total{hasRTT="true"}[1m])) / sum(rate(netobserv_agent_flows_enrichment_total[1m]))`

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -2,6 +2,7 @@ package metrics
 
 import (
 	"errors"
+	"strconv"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
@@ -118,10 +119,20 @@ var (
 	)
 	errorsCounter = defineMetric(
 		"errors_total",
-		"errors counter",
+		"Errors counter",
 		TypeCounter,
 		"component",
 		"error",
+	)
+	flowEnrichmentCounterCounter = defineMetric(
+		"flows_enrichment_total",
+		"Statistics on flows enrichment",
+		TypeCounter,
+		"hasDNS",
+		"hasRTT",
+		"hasDrops",
+		"hasNetEvents",
+		"hasXlat",
 	)
 )
 
@@ -154,6 +165,7 @@ type Metrics struct {
 	NetworkEventsCounter  *EvictionCounter
 	BufferSizeGauge       *BufferSizeGauge
 	Errors                *ErrorCounter
+	FlowEnrichmentCounter *FlowEnrichmentCounter
 }
 
 func NewMetrics(settings *Settings) *Metrics {
@@ -168,6 +180,7 @@ func NewMetrics(settings *Settings) *Metrics {
 	m.NetworkEventsCounter = &EvictionCounter{vec: m.NewCounterVec(&networkEvents)}
 	m.BufferSizeGauge = &BufferSizeGauge{vec: m.NewGaugeVec(&bufferSize)}
 	m.Errors = &ErrorCounter{vec: m.NewCounterVec(&errorsCounter)}
+	m.FlowEnrichmentCounter = &FlowEnrichmentCounter{vec: m.NewCounterVec(&flowEnrichmentCounterCounter)}
 	return m
 }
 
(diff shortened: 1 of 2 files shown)


pkg/metrics/metrics.go
```diff
@@ -118,11 +119,20 @@ var (
...
 		"error",
 	)
+	flowEnrichmentCounterCounter = defineMetric(
+		"flows_enrichment_total",
+		"Statistics on flows enrichment",
+		TypeCounter,
+		"hasDNS",
+		"hasRTT",
+		"hasDrops",
+		"hasNetEvents",
```
msherif1234, line 134: we need `hasXlat` counter as well

jotak: done

pkg/tracer/tracer.go
```diff
@@ -905,6 +906,20 @@ func (m *FlowFetcher) LookupAndDeleteMap(met *metrics.Metrics) map[ebpf.BpfFlowI
...
 }
 
+func (m *FlowFetcher) increaseEnrichmentStats(met *metrics.Metrics, flow *model.BpfFlowContent) {
+	if flow.AdditionalMetrics != nil {
+		met.FlowEnrichmentCounter.Increase(
+			flow.AdditionalMetrics.DnsRecord.Id != 0,
+			flow.AdditionalMetrics.FlowRtt != 0,
+			flow.AdditionalMetrics.PktDrops.Packets != 0,
+			flow.AdditionalMetrics.NetworkEventsIdx != 0,
+			flow.AdditionalMetrics.TranslatedFlow.ZoneId != 0,
```
msherif1234, line 916: zoneid of 0 is valid

jotak: hmm :/
I wanted to avoid something more costly, so I guess I need to check for empty IPs ...

memodi:
>
Or, to have the proportion of flows with RTT for instance:
100 * sum(rate(netobserv_agent_flows_enrichment_total{hasRTT="true"}[1m])) / sum(rate(netobserv_agent_flows_enrichment_total[1m]))

@jotak - for such metrics , I think it would be useful to have queries that shows delta between the time period, for example, with the traffic being constant, if some feature enrichment stats drops heavily in a 5 min interval compared to previous 5min interval. Is it possible to have example of such queries?

pkg/metrics/metrics.go
```diff
@@ -118,11 +119,21 @@ var (
 	)
 	errorsCounter = defineMetric(
 		"errors_total",
-		"errors counter",
+		"Errors counter",
```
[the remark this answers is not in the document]
jotak: Just for consistency, all other help texts are capitalized .E.g "Sampling rate" just above

pkg/tracer/tracer.go
```diff
@@ -929,6 +930,20 @@ func (m *FlowFetcher) LookupAndDeleteMap(met *metrics.Metrics) map[ebpf.BpfFlowI
 	return flows
 }
 
+func (m *FlowFetcher) increaseEnrichmentStats(met *metrics.Metrics, flow *model.BpfFlowContent) {
+	if flow.AdditionalMetrics != nil {
+		met.FlowEnrichmentCounter.Increase(
+			flow.AdditionalMetrics.DnsRecord.Id != 0,
+			flow.AdditionalMetrics.FlowRtt != 0,
+			flow.AdditionalMetrics.PktDrops.Packets != 0,
+			flow.AdditionalMetrics.NetworkEventsIdx != 0,
```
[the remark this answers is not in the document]
jotak: oh, right
`!model.AllZerosMetaData(flow.AdditionalMetrics.NetworkEvents[0])` would be better I guess

jotak:
> @jotak - for such metrics , I think it would be useful to have queries that shows delta between the time period, for example, with the traffic being constant, if some feature enrichment stats drops heavily in a 5 min interval compared to previous 5min interval. Is it possible to have example of such queries?

First time I'm doing this, but yeah, sounds like it's doable, e.g:

Here's a rate function:
```
 sum(rate(netobserv_agent_flows_enrichment_total{hasRTT="true"}[1m]))
```

Capture d’écran du 2025-01-14 13-09-59

Here's the rate derivative (second derivative) :
```
deriv(sum(rate(netobserv_agent_flows_enrichment_total{hasRTT="true"}[1m]))[2m:15s])
```

Capture d’écran du 2025-01-14 13-10-07

It shows the derivative for changes in the rate in the last 2 minutes. In other words, when this metric is not close to 0, it means there has been changes in the last 2 minutes.

PR #487, opened 2024-12-14
WIP: Initial intg with UDN interface mapping api

msherif1234:
## Dependencies

n/a

## Checklist


jotak (commented):
just one question/remark, other than that lgtm

pkg/pbflow/proto.go
```diff
@@ -91,10 +91,27 @@ func FlowToPB(fr *model.Record, s *ovnobserv.SampleDecoder) *Record {
...
+			Udn:       "",
+		}
+		if s != nil {
+			m, err := s.GetInterfaceUDNs()
+			if err == nil {
+				if v, ok := m[entry.Interface]; ok {
+					if v != "" {
+						entry.Udn = v
+					} else {
+						entry.Udn = "default-network"
```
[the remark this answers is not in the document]
msherif1234: I don't think it will ever conflict with udn tag it uses name/namespace syntax but I will double check with ovnk team to be sure

pkg/exporter/converters_test.go
```diff
@@ -56,6 +56,7 @@ func TestConversions(t *testing.T) {
 					},
 				},
 				Interfaces:    []model.IntfDir{model.NewIntfDir("eth0", model.DirectionEgress)},
+				UdnList:       []string{""},
```
jotak, line 59: could you add some non-empty data in some of the tests?

msherif1234: sure will do

pkg/pbflow/proto.go
```diff
@@ -90,10 +90,27 @@ func FlowToPB(fr *model.Record, s *ovnobserv.SampleDecoder) *Record {
...
 	pbflowRecord.DupList = make([]*DupMapEntry, 0)
 	for _, intf := range fr.Interfaces {
-		pbflowRecord.DupList = append(pbflowRecord.DupList, &DupMapEntry{
+		entry := DupMapEntry{
 			Interface: intf.Interface,
 			Direction: Direction(intf.Direction),
-		})
+			Udn:       "",
+		}
+		if s != nil {
```
jotak, line 98: Doing this in protobuf decode will cause the same issue as what I mentioned here: https://github.com/netobserv/netobserv-ebpf-agent/pull/286#pullrequestreview-2230254996
.. and which made us having this bug: https://issues.redhat.com/browse/NETOBSERV-2009
I think it doesn't make sense to have any enrichment set in protobuf decode. There should be in a common path, not something that may end up not being invoked. IMO we should fix that in a first place rather than adding technical debt.

pkg/exporter/converters_test.go
msherif1234: well actually this will require mock to ovnk decoded to return back strings

pkg/pbflow/proto.go
msherif1234: I think adding new pipeline might be very risk at this stage given the unstability e have can this be deferred to 1.9 ?

jotak: An option could be to create a new stage (like in [this area](https://github.com/netobserv/netobserv-ebpf-agent/blob/ca5b9445508a8befa8fe65eefd8f29d460733f79/pkg/agent/agent.go#L504-L515)), or perhaps inject some enrichment callback directly in `NewRecord` [like it's done for interfaceNamer](https://github.com/netobserv/netobserv-ebpf-agent/blob/main/pkg/model/record.go#L89), wdyt?

jotak: We should fix NETOBSERV-2009 in 1.8 so if we fix it for events we could surely fix it as well for udns?

msherif1234: we still in TP for network events so no urgency for CLI support I think for 1.8 I will try to explode the new stage option

pkg/pbflow/proto.go
```diff
@@ -90,10 +90,27 @@ func FlowToPB(fr *model.Record, s *ovnobserv.SampleDecoder) *Record {
...
+			m, err := s.GetInterfaceUDNs()
```
jotak, line 99: Looking at the ovn code, it seems like `s.GetInterfaceUDNs()` doesn't just returned some cached thing, it calls `d.ovsdbClient.List(context.Background(), &ifaces)` under the hood, which may involve some i/o ? That would be nice to double check. I think it's better _at the very least_ to move this call out of the interface for loop, but even better, it could be fetched just once for the whole eviction loop ?

msherif1234: ovsdb is a unix socket to ovs DB `/var/run/openvswitch/db.sock` and I believe its cache based too, doing once per evicttion won't work as we might miss tags as pods comes and goes

jotak: even if there's a cache doesn't mean there will never be i/o - it sounds risky to have that in the agent data path
... checking with the ovn folks what it really is about

pkg/model/record.go
```diff
@@ -8,8 +8,10 @@ import (
 	"reflect"
 	"time"
 
-	"github.com/netobserv/flowlogs-pipeline/pkg/config"
 	"github.com/netobserv/netobserv-ebpf-agent/pkg/ebpf"
+
+	ovnmodel "github.com/ovn-org/ovn-kubernetes/go-controller/observability-lib/model"
+	ovnobserv "github.com/ovn-org/ovn-kubernetes/go-controller/observability-lib/sampledecoder"
```
jotak, line 14: I don't want to block this PR further, but I think we should get rid of these dependencies in our model package
So, no blocker - I suggest to add that to a follow-up task - possibly for 1.9.

An option could be to pass an interface, or callbacks, instead of the actual sampledecoder.
(Another option which we briefly discussed is to add a new stage instead of doing this decoration directly in `Record`, but I'm not sure it's any better)

pkg/model/record.go
```diff
@@ -97,16 +107,72 @@ func NewRecord(
...
+		return IntfDirUdn{Interface: intf, Direction: dir, Udn: ""}
+	}
+
+	// Load UDN cache if empty
+	if len(udnsCache) == 0 {
+		m, err := s.GetInterfaceUDNs()
+		if err != nil {
+			return IntfDirUdn{Interface: intf, Direction: dir, Udn: ""}
+		}
+		udnsCache = m
```
jotak, line 162: this cache is never updated? I can add that also as a follow-up if you want

pkg/pbflow/proto.go
```diff
@@ -110,43 +108,12 @@ func FlowToPB(fr *model.Record, s *ovnobserv.SampleDecoder) *Record {
...
-						pbflowRecord.NetworkEventsMetadata = append(pbflowRecord.NetworkEventsMetadata, &pbEvent)
-						seen[mdStr] = true
-					}
-				} else {
-					protoLog.Errorf("unable to decode Network events cookie: %v", err)
-				}
-			}
-		}
+
+	pbflowRecord.NetworkEventsMetadata = make([]*NetworkEvent, 0)
```
jotak, line 112: nit / follow-up: avoid creating NetworkEventsMetadata if fr.NetworkMonitorEventsMD is empty?

pkg/decode/decode_protobuf.go
```diff
@@ -69,12 +69,19 @@ func RecordToMap(fr *model.Record) config.GenericMap {
 
 	var directions []int
 	var interfaces []string
+	var udns []string
 	for _, intf := range fr.Interfaces {
 		directions = append(directions, intf.Direction)
 		interfaces = append(interfaces, intf.Interface)
+		if intf.Udn != "" {
```
jotak, line 76: shouldn't we append udns here, regardless of it's presence? It depends how this data is processed in the console plugin. If the udn array is correlated with the interface array, the indexes in udn array must match the one from the interfaces. Not totally sure about that.

msherif1234: in the console side they are two separate list? u mean when user correlate between the two in the console ui they won't be in sync so maybe its better to show it regardless I will remove this check

pkg/model/record.go
msherif1234: yeah we can see how to clean this further in 1.9 I briefly looked at new stage but I quickly realized it adds some overhead and wasn't sure how to correctly wire things

pkg/model/record.go
msherif1234: well the caches is reset every eviction then read from ovnk after that its just ready only no update is needed during the eviction processing

pkg/decode/decode_protobuf.go
jotak: honestly I don't remember how it's one in the console plugin ... For interfaces and direction I know that we have a column that shows both "Interfaces and Directions" so the 2 arrays are correlated; but did we do something similar for udn, I'm not sure

pkg/model/record.go
jotak: oh yes I didn't noticed in NewRecord

jotak (approved):
/lgtm

PR #488, opened 2024-12-16
optimize filter sampling logic when filtering not enabled

msherif1234:
## Description

optimize logic when filtering not enabled to avoid doing packet parsing when its not needed
## Dependencies

n/a

## Checklist


jotak:
Thanks, I think it fixes most of the regression but there are still a case not covered: if users previously had a filter and have global sampling, they would still see a performance regression.
IMO to fix the regression entirely we would need to introduce a global config like `has_filter_sampling`, defined from user-space and passed to the bpf prog, that is set to one if there is at least one filter defined with sampling.

msherif1234:
> Thanks, I think it fixes most of the regression but there are still a case not covered: if users previously had a filter and have global sampling, they would still see a performance regression. IMO to fix the regression entirely we would need to introduce a global config like `has_filter_sampling`, defined from user-space and passed to the bpf prog, that is set to one if there is at least one filter defined with sampling.

if u enable filtering that should by itself optimize things I think we can defer this till we better know how the filter smapling will be used we might need to have default value also in case its not set ?

jotak:
@msherif1234 
Even though filtering reduces the resource footprint, that's still a regression that we would let go in while we could avoid it quite easily

jotak:
@msherif1234 here's my suggestion to fix this entirely: https://github.com/msherif1234/netobserv-ebpf-agent/pull/1

bpf/utils.h
```diff
@@ -211,7 +218,7 @@ static inline bool check_and_do_flow_filtering(flow_id *id, u16 flags, u32 drop_
             // we have no matching rules so we update global counter for flows that are not matched by any rule
             increase_counter(FILTER_NOMATCH);
             // we have accept rule but no match so we can't let mismatched flows in the hashmap table.
-            if (action == ACCEPT || action == MAX_FILTER_ACTIONS) {
+            if (action == ACCEPT) {
```
jotak, line 221: I guess I see why you removed this, I think it fixes an issue but introduces a new one. In case of no CIDR match, action is neither ACCEPT nor REJECT and previously we would view that as a DROP; now you change that to be a KEEP. I think neither of those options are valid, we need to inspect the filters to decide whether it should be DROP or KEEP.
If users have defined any ACCEPT rules, then a no-match should be DROP
If users have defined any REJECT rules, then a no-match should be KEEP

... and if users have defined a mix of ACCEPT and REJECT, I think it just doesn't make sense because we don't know what to say when no rule is matched?

So at the end of the day, I think it's an API issue: we shouldn't allow to set ACCEPT/REJECT action _per rule_, it should be a single setting for all rules.

msherif1234: we should be the one define what is the default behavior when no CIDR match happens which is when we get the special action value, in that case I feel keeping the default as before to disallow is more accurate , nothing wrong with the API IMO, we just need to define what is the nomatch action behavior

msherif1234:
> @msherif1234 Even though filtering reduces the resource footprint, that's still a regression that we would let go in while we could avoid it quite easily
> 
> > till we better know how the filter smapling will be used we might need to have default value also in case its not set
> 
> The regression that remains is without using filter sampling, let's say someone was filtering for only TCP traffic and applied a global sampling, this too would be concerned with a performance regression

I understand but my point is when filtering is on u see same perf impact regardless if u  hvae sampoling action or not its not worth optimizing this case IMO

bpf/utils.h
jotak: I don't agree, there's no point having ALLOW by default if you define ALLOW rules, and there's no point having DENY by default if you define DENY rule. It's an API problem.
If we keep as it is today, then there's a whole range of things you can't express, e.g:

```
Filter: { 1.2.3.4/32: REJECT }
```

You would expect a flow like that to be kept: src=6.6.6.6, dst=8.8.8.8
But here it would be dropped because action == MAX_FILTER_ACTIONS

jotak:
> I understand but my point is when filtering is on u see same perf impact regardless if u hvae sampoling action or not its not worth optimizing this case IMO

So we would be introducing a perf regression on purpose, just so it aligns with a case that has worse performance? I'm not sure to agree with that argument... And the optimization is not super complicated.

msherif1234:
> > I understand but my point is when filtering is on u see same perf impact regardless if u hvae sampoling action or not its not worth optimizing this case IMO
> 
> So we would be introducing a perf regression on purpose, just so it aligns with a case that has worse performance? I'm not sure to agree with that argument... And the optimization is not super complicated.

it seems overkill to customize for special case IMO what if people endup using sampling all the time when they filter they might complain why filtering with smapling has different cpu than with sampling , idk seems too much IMO but not totally against it regardless if its easy or hard to do

jotak:
I don't think that people doing filtering + global sampling are a corner case to ignore. Doing simple filtering with global sampling is precisely meant to reduce resource consumption - the new filtered sampling is a nice feature to fine-tune what you sample, but we can't assume everybody will use that.
I just did a test with global sampling 50 and filter only TCP traffic: before multi-filter PR, CPU reported by bpftop was around 1.95% under stress. With multi-filter + this PR, it climbs to 5.76% => that's almost a 200% increase


Before:
Capture d’écran du 2024-12-18 15-57-28

With this PR:
Capture d’écran du 2024-12-18 16-03-45

bpf/utils.h
msherif1234: yeah u need to add Allow 0.0.0.0/0

msherif1234:
> I don't think that people doing filtering + global sampling are a corner case to ignore. Doing simple filtering with global sampling is precisely meant to reduce resource consumption - the new filtered sampling is a nice feature to fine-tune what you sample, but we can't assume everybody will use that. I just did a test with global sampling 50 and filter only TCP traffic: before multi-filter PR, CPU reported by bpftop was around 1.95% under stress. With multi-filter + this PR, it climbs to 5.76% => that's almost a 200% increase
> 
> Before: Capture d’écran du 2024-12-18 15-57-28
> 
> With this PR: Capture d’écran du 2024-12-18 16-03-45

I am fine proceeding with the suggested optimization tell we better know how users will endup using this since u have a ready PR we can get mine in then u can followup with ur change WDYT ?

bpf/utils.h
jotak: When you add Allow 0.0.0.0/0, you fall into another problem: since the src_ip is checked before the dst_ip, this global-allow will take precedence on a flow like `src=6.6.6.6, dst=1.2.3.4` even though there is the reject rule for that IP that in theory should take precedence.
Maybe then you need to always run both lookups for source and dest, and not check dest only when src has no match.

I think we should be more clear on what to expect with this feature, I've started a doc for acceptance criteria: https://docs.google.com/document/d/1YUVcpCdYxRACRVrqEOwir6upMyla5RvvUsZmgP4WvIs/edit?tab=t.0 - I think there's already a couple of issues - things that are not covered

jotak:
@msherif1234 can you review this commit https://github.com/netobserv/netobserv-ebpf-agent/pull/488/commits/ec403e4b25edb36b4ae83720e0127fda7bd0d25f ?

pkg/tracer/tracer.go
```diff
@@ -881,25 +879,37 @@ func (m *FlowFetcher) LookupAndDeleteMap(met *metrics.Metrics) map[ebpf.BpfFlowI
...
+	addtlIterator := m.objects.AdditionalFlowMetrics.Iterate()
+	for addtlIterator.Next(&id, &additionalMetrics) {
+		ids = append(ids, id)
+	}
+
+	countAdditional := 0
+	for _, id := range ids {
+		countAdditional++
+		if err := m.objects.AdditionalFlowMetrics.LookupAndDelete(&id, &additionalMetrics); err != nil {
+			log.WithError(err).WithField("flowId", id).Warnf("couldn't lookup/delete additional metrics entry")
```
msherif1234, line 897: why the error handling here is different than the other map in fact this more identical to old code so it should be similar no ? specially the check for `i===0` ?

bpf/types.h
```diff
@@ -106,6 +106,9 @@ typedef struct flow_metrics_t {
 const struct flow_metrics_t *unused2 __attribute__((unused));
 
 typedef struct additional_metrics_t {
+    u64 start_mono_time_ts;
```
msherif1234, line 109: do we really need this for all features or only the drop case if drop case only we might bundle this in the drop struct ? WDYT ? might be cleaner to have it for all and more future proof just in case not sure

jotak: yes I think it's safer to do it for all. It's actually a good thing that the drop=true filter revealed this issue, because it certainly can happen also in other situations, even if that's exceptional. There certainly can be some race condition that makes flows "orphan" in the additional map, without counterpart in main flow map, so we need to fetch &delete them separately if we don't want to risk having them in maps forever. And those "orphan" flows need the timestamp & eth_protocol to be correctly processed

msherif1234:
/ok-to-test
cc'd @memodi this PR will fix drop filter issue u have seen let me know if i can move it back to review

jotak:
> I am fine proceeding with the suggested optimization tell we better know how users will endup using this since u have a ready PR we can get mine in then u can followup with ur change WDYT ?
Ok

bpf/flows.c
```diff
@@ -79,6 +80,17 @@ static inline void update_dns(additional_metrics *extra_metrics, pkt_info *pkt,
...
+    u32 filter_sampling = 0;
+
+    if (!is_filter_enabled()) {
+        if (sampling > 1 && (bpf_get_prandom_u32() % sampling) != 0) {
+            do_sampling = 0;
+            return TC_ACT_OK;
+        }
+        filter_sampling = sampling;
+        do_sampling = 1;
+    }
```
stleerh, line 92: What should `do_sampling` be set to if filter is enabled?  It is currently left unchanged.

bpf/flows.c
```diff
@@ -103,21 +115,21 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
...
+        }
+        // If sampling is defined, will only parse 1 out of "sampling" flows
+        if (filter_sampling > 1 && (bpf_get_prandom_u32() % filter_sampling) != 0) {
+            do_sampling = 0;
+            return TC_ACT_OK;
+        }
+        do_sampling = 1;
+        if (skip) {
+            return TC_ACT_OK;
+        }
```
stleerh, line 131: This is related to the same question on what `do_sampling` should be set to since this is part of the same function.

bpf/flows.c
msherif1234: do_sampling is just bool either 0 or 1 sampling rate is what prandom uses

stleerh: Prior to this code change, `do_sampling` was always set to 0 or 1.  Now it can be left unchanged if `is_filter_enabled()` is false.  There are a few places in other parts of the code that does something depending on the state of `do_sampling`.  See: [https://github.com/search?q=repo%3Anetobserv/netobserv-ebpf-agent%20do_sampling&type=code](https://github.com/search?q=repo%3Anetobserv/netobserv-ebpf-agent%20do_sampling&type=code).

jotak: @stleerh I think all code paths are covered to set a value on do_sampling:
- if filters aren't enabled (line 85), do_sampling is set to either 0 or 1
- if filters are enabled (line 118), do_sampling is also set to either 0 or 1

The execution between line 85 and 118 are just IP/TCP headers extraction, they won't involve checking do_sampling, so that should be fine

stleerh: Okay, I didn't see that line 85 is not the same as line 118.

/lgtm

PR #494, opened 2024-12-18
Memset dns id to 0

jotak:
That's a regression following the [map split PR](https://github.com/netobserv/netobserv-ebpf-agent/pull/469), which removed `packed` attributes; now that map keys aren't packed anymore, it becomes critical to initialize them: failing to do so results in having many map misses on lookup due to pad fields having random values.


msherif1234:
pls add in the description this was a regression caused by https://github.com/netobserv/netobserv-ebpf-agent/pull/469 for better tracking, I checked other hooks they seem fine and initialize their key.
/lgtm

PR #496, opened 2024-12-20
Restore performances in filtering case

jotak:
Follow-up on https://github.com/netobserv/netobserv-ebpf-agent/pull/488

Return early in case of filters being set without any conditional sampling.
As discussed [here](https://github.com/netobserv/netobserv-ebpf-agent/pull/488#issuecomment-2551585544) , some use cases would see a non negligible performance drop if we don't restore an optimal early return.


pkg/tracer/tracer.go
```diff
@@ -162,8 +163,15 @@ func NewFlowFetcher(cfg *FlowFetcherConfig) (*FlowFetcher, error) {
 		}
 
 		enableFlowFiltering := 0
+		hasFilterSampling := uint8(0)
 		if cfg.EnableFlowFilter {
 			enableFlowFiltering = 1
+			for _, f := range cfg.FilterConfig {
```
msherif1234, line 169: nit: we can have this as function in flow_filter.go and call it or even move the flow_filter processing up and let it return the flag not urgent can be done later

msherif1234:
remember to delete the new const from the spec in `NewPacketFetcher`

pkg/tracer/tracer.go
jotak: hmm yes ... while doing that I realize it's not handled in the bpfman case .. I need to duplicate that logic in the operator, for bpfman, right?

jotak:
> remember to delete the new const from the spec in NewPacketFetcher
@msherif1234  actually, shouldn't I add the same logic in the PacketFetcher code path ?

msherif1234:
> > remember to delete the new const from the spec in NewPacketFetcher
> 
> @msherif1234 actually, shouldn't I add the same logic in the PacketFetcher code path ?

packetfiltering has sampling on by default and it doesn't support sampling override

PR #508, opened 2025-01-14
NETOBSERV-2055: Xlat ICMP code is redundant with icmp code field so removing it

msherif1234:
## Description

check the protocol to set the appropriate fields in xlat record
## Dependencies

n/a

## Checklist


msherif1234:
cc'd @Amoghrd 
/ok-to-test

Amoghrd:
@msherif1234 It still shows up as 0 instead of n/a. I tried both ping and curl to the service IP from the client pod and resulting flows are seen below
Screenshot 2025-01-14 at 3 32 23 PM

Flows with ICMP protocol have ICMP code 0 and Xlat ICMP code 0.
But flows with TCP protocol have ICMP code n/a and Xlat ICMP code 0 which is incorrect.

And could the JSON for Xlat ICMP filed be updated from `XlatIcmpId` to `XlatIcmpCode` to remain consistent with default field names
Screenshot 2025-01-14 at 3 35 09 PM

msherif1234:
I added check for 0 too but we have to update flp 1st to be able to test this

bpf/utils.h
```diff
@@ -358,4 +358,14 @@ static inline void fill_in_others_protocol(flow_id *id, u8 protocol) {
     id->transport_protocol = protocol;
 }
 
+static inline bool is_transport_protocol(u8 protocol) {
+    switch (protocol) {
+    case IPPROTO_TCP:
+    case IPPROTO_UDP:
+    case IPPROTO_SCTP:
```
jotak, line 365: Looking into the linux headers / `nf_conntrack_man_proto`, sounds like DCCP can also have port?

msherif1234: but we don't support or parse that protocol in openshift

jotak: yeah never mind .. no point doing it in xlat if we don't do it in the main path anyway

PR #509, opened 2025-01-15
NETOBSERV-2075: Simplifying observed interfaces - moving it to main map

jotak:
Observed interfaces is a structure managed from the TC hook, so it is elligible for being written in the main map instead of the additional metrics map. Moreover, it is very frequently created (the vast majority of flows being observed from at least two interfaces), so having it in the additional maps sort of defeats one of its purpose, that was allowing to reduce the memory used by separating commonly flow data from more uncommon data.

In other words, when this is set into the addtional map, the size map becomes similar to main map, whereas if removed the size map can be shrinked to a much lower number.

Also:
- Increase Max Interfaces
- ignore direction for duplicate interface checks (tradeoff to minimize loss of data in case of max reached)
- Discard non IP flows (as they all end up in the same map bucket, they make the max interface quickly reached)

## Dependencies

n/a

## Checklist


bpf/flows.c
```diff
@@ -155,29 +155,7 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
...
-                    extra_metrics =
-                        (additional_metrics *)bpf_map_lookup_elem(&additional_flow_metrics, &id);
-                    if (extra_metrics != NULL) {
-                        add_observed_intf(extra_metrics, skb->ifindex, direction);
-                    }
-                } else if (ret != 0 && trace_messages) {
-                    bpf_printk("error creating new observed_intf: %d\n", ret);
-                }
-            }
+            add_observed_intf(aggregate_flow, skb->ifindex, direction);
```
jotak, line 158: need to acquire lock here

msherif1234: make interface update part of `update_existing_flow()`

jotak: done, but I still take the lock in add_observed_intf for 2 reasons:
- we can't call a function while having a lock (so for instance the call to `increase_counter` isn't allowed)
- we can't read map value while having the lock, so all the for-loop to check for existing intf+direction isn't allowed during locking
In other words we need to keep lock as short time as possible

bpf/types.h
```diff
@@ -104,6 +104,9 @@ typedef struct flow_metrics_t {
     // https://chromium.googlesource.com/chromiumos/docs/+/master/constants/errnos.md
     u8 errno;
     u8 dscp;
+    u8 nb_observed_intf;
+    u8 observed_direction[MAX_OBSERVED_INTERFACES];
+    u32 observed_intf[MAX_OBSERVED_INTERFACES];
```
msherif1234, line 109: why u changed this instead of having array of struct ?

bpf/flows.c
```diff
@@ -55,16 +52,47 @@
...
-    aggregate_flow->dscp = pkt->dscp;
-    aggregate_flow->sampling = sampling;
-    bpf_spin_unlock(&aggregate_flow->lock);
+static inline void add_observed_intf(flow_metrics *value, pkt_info *pkt, u32 if_index,
+                                     u8 direction) {
+    u8 nb_observed_intf = value->nb_observed_intf;
+    if (nb_observed_intf < MAX_OBSERVED_INTERFACES) {
+        for (u8 i = 0; i < nb_observed_intf; i++) {
+            if (value->observed_intf[i] == if_index) {
+                // Interface already seen -> skip (we don't check for direction to avoid reaching the max too frequently, as a tradeoff)
```
msherif1234, line 61: what is the impact of ignoring direction ?

bpf/flows.c
```diff
@@ -55,16 +52,47 @@
...
+    } else {
+        increase_counter(OBSERVED_INTF_MISSED);
+        BPF_PRINTK("observed interface missed (array capacity reached) for ifindex %d\n", if_index);
+    }
+}
+
+static __always_inline void update_existing_flow(flow_metrics *aggregate_flow, pkt_info *pkt,
+                                                 u64 len, u32 sampling, u32 if_index,
+                                                 u8 direction) {
+    // Count only packets seen from the same interface as previously to avoid duplicate counts
```
msherif1234, line 82: I think u need to grap the lock here and free after the else and remove it from add_observred_intf to make sure not reading any value while it can be updated from another thead on different core

bpf/types.h
jotak: It's padding optimization. With a struct there was a 7 bytes padding per struct instance, so 42 bytes in total. Separated arrays reduce it to a minimum.

bpf/flows.c
jotak: We discussed that in slack. The result is not marking twice the same interface when it has a different direction. E.g. if a flow shows eth0/egress then eth0/ingress, only eth0/egress is going to be stored.
We may document that somewhere as a limitation ...

jotak: as said in slack an alternative could be to introduce a new direction enum "BOTH", if we think it's important. It would take both of both worlds, ie. avoid reaching the max interface and still telling everything about what was observed. But it needs changes in the console plugin for display. I can do that as a follow-up if we think it's important

bpf/flows.c
jotak: this is not possible, that's what I said here https://github.com/netobserv/netobserv-ebpf-agent/pull/509#discussion_r1918300302
- when lock is acquired, no call function is allowed. Currently there's two call functions: `add_observed_intf` and `increase_counter`
- reading map value isn't allowed while lock is acquired. For instance, the line `if (value->observed_intf[i] == if_index)` in `add_observed_intf` generates an error

If we really wanted to, we could work that around, inlining everything and creating temporary copies of values, but I think it would just make it worse.
The rule of thumb is to keep the lock as little time as possible, which is what I think I'm doing here. If the value is read from another thread before the lock is acquired, I don't think it's a problem, it's just going to get the value before it's updated, right?

msherif1234: even if u use always_inline, the risk I see assume one thread added one interface and inc the count while another thread already looped over the prev counter and we end up overwritting it, is the error compiler err or verification err can u share ?

jotak: It's in the verifier, on startup:
Without inlining:
```
...ES) {\n\t1528: (25) if r1 > 0x5 goto pc+100 1629: frame1: R1=scalar(umin=6,umax=255,var_off=(0x0; 0xff)) R3=0 R5=1 R6=map_value(id=5,off=48,ks=40,vs=96,imm=0) R7=scalar(umax=4294967295,var_off=(0x0; 0xffffffff)) R8=0 R9=map_value(id=5,off=0,ks=40,vs=96,imm=0) R10=fp0
 fp-32=????mm?? fp-40=00000000 fp-48= fp-56=00000000 fp-64=00000000 fp-72=mmmmmmmm fp-80=fp fp-88=mmmmmmmm fp-96=mmmmmmmm fp-104=mmmmmmmm fp-112=mmmmmmmm fp-120=mmmmmmmm fp-272=ctx fp-280=pkt fp-288=00000000 fp-296= fp-304=1 fp-312=00000000\n\t; bpf_printk(\"couldn't reserve space in the ringbuf. Dropping flow\");\n\t1629: (b7) r1 = 9
                     ; frame1: R1_w=9\n\t1630: (63) *(u32 *)(r10 -264) = r1    ; frame1: R1_w=9 R10=fp0 fp-264=9\n\t1631: (b7) r8 = 1                     ; frame1: R8_w=1\n\t; u32 initVal = 1;\n\t1632: (63) *(u32 *)(r10 -28) = r8     ; frame1: R8_w=1 R10=fp0 fp-32=mmmmmm??\n\t1633: (bf) r2 = r10
                   ; frame1: R2_w=fp0 R10=fp0\n\t;\n\t1634: (07) r2 += -264                 ; frame1: R2_w=fp-264\n\t; error_counter_p = bpf_map_lookup_elem(&global_counters, &key);\n\t1635: (18) r1 = 0xffff8b2008172c00    ; frame1: R1_w=map_ptr(off=0,ks=4,vs=4,imm=0)\n
\t1637: (85) call bpf_map_lookup_elem#1\n\tfunction calls are not allowed while holding a lock\n\tprocessed 482 insns (limit 1000000) max_states_per_insn 2 total_states 29 peak_states 29 mark_read 15" component=ebpf.FlowFetcher
time="2025-01-21T12:34:34Z" level=fatal msg="can't instantiate NetObserv eBPF Agent" error="loading and assigning BPF objects: field TcEgressFlowParse: program tc_egress_flow_parse: load program: invalid argument: function calls are not allowed while holding a lock (280 line(s) omitted)"
```
=> **function calls are not allowed while holding a lock**

So we need to refactor and/or use always_inline ... That's actually tricky because `add_observed_intf` includes `increase_counter` which contains helper calls (bpf_map_lookup_elem) that cannot be inlined, so in any case that would require some refactoring.

I mentioned a second error that I saw last time, but I don't reproduce it today, so not sure what it was exactly....

So ... let me try to clean up all that and see how that works...

msherif1234: u can make add_intf return and error and get the counter outside at the caller to avoid nested functions

bpf/flows.c
jotak: It's mostly theoretical, I don't think it involve any real loss in user experience - in fact, when I was seeing this the other day, I'm now pretty sure it was due to the non-IP flows that we're [not capturing anymore](https://github.com/netobserv/netobserv-ebpf-agent/pull/509/commits/938e96dce3cc3ab1b654a0c0aa855228b99d15eb).

Anyway, it's not a big deal to manage "both" directions so I've added that in the last commit. And on console side: https://github.com/netobserv/network-observability-console-plugin/pull/695 (I generated fake data to be able to see it in console)

bpf/flows.c
```diff
@@ -55,16 +52,53 @@
...
+    if (value->nb_observed_intf >= MAX_OBSERVED_INTERFACES) {
+        BPF_PRINTK("observed interface missed (array capacity reached) for ifindex %d\n", if_index);
+        return 1;
+    }
+    for (u8 i = 0; i < value->nb_observed_intf; i++) {
+        if (value->observed_intf[i] == if_index) {
+            if (value->observed_direction[i] != direction &&
+                value->observed_direction[i] != OBSERVED_DIRECTION_BOTH) {
+                // Same interface seen on a different direction => mark as both directions
+                value->observed_direction[i] = OBSERVED_DIRECTION_BOTH;
```
msherif1234, line 67: won't this case issues in flp generating the other flp direction field ?

jotak: FLP doesn't use the interface direction to figure out the node direction, it does that only by comparing agent IP versus src/dest node IP
I don't think there's anything that reads the IfDirections field, except in the console for display...

msherif1234: what is the impact of using direction both when we filter flows on direction ?

bpf/types.h
```diff
@@ -66,7 +66,8 @@ typedef __u64 u64;
 #define MAX_FILTER_ENTRIES 16
 #define MAX_EVENT_MD 8
 #define MAX_NETWORK_EVENTS 4
-#define MAX_OBSERVED_INTERFACES 4
+#define MAX_OBSERVED_INTERFACES 6
+#define OBSERVED_DIRECTION_BOTH 3
```
msherif1234, line 70: we have enum used to define possible directions https://github.com/netobserv/netobserv-ebpf-agent/blob/main/bpf/types.h#L75
so u can add both there and move MAX to 3

jotak: I didn't want to change the enum to not mess up anything with the filters, which uses that enum. This value is very specific to the "observed direction" thing

memodi:
@jotak It's hard to tell deterministically whether we're seeing improvements because these errors are still expected:

Screenshot 2025-01-23 at 1 24 09 PM

in above I am using query: `sum(rate(netobserv_agent_errors_total[1m])*60) by (error)`

jotak:
@memodi weird .. when I tested it, they almost disappeared. There might have been something regressed since my first commits, let me re-investigate

bpf/flows.c
```diff
@@ -55,16 +52,60 @@
...
+        aggregate_flow->flags |= pkt->flags;
+        maxReached = add_observed_intf(aggregate_flow, pkt, if_index, direction);
+    }
     bpf_spin_unlock(&aggregate_flow->lock);
+    if (maxReached > 0) {
+        BPF_PRINTK("observed interface missed (array capacity reached); ifindex=%d, eth_type=%d, "
+                   "proto=%d, sport=%d, dport=%d\n",
+                   if_index, aggregate_flow->eth_protocol, pkt->id->transport_protocol,
+                   pkt->id->src_port, pkt->id->dst_port);
+        if (pkt->id->transport_protocol != 0 &&
```
msherif1234, line 103: icmpv4/v6 won't work with this check ?

jotak: yes, anything with both ports = 0 won't report that as an error, it will limit the interfaces silently - I'm trying to place the cursor between making this issue too visible and too hidden

msherif1234: wouldn't protocol != 0 should be good enough ?

jotak: done
quick tests on my cluster show that it's sufficient indeed, we'll see if that's good enough for @memodi too

memodi:
thanks @jotak , with latest image seeing huge improvements, I'll also make change in perf testing to only look for {severity!=low} agent errors

image

PR #525, opened 2025-01-24
fix linter issue in latest code

msherif1234:
## Description

```sh
pkg/model/flow_content.go:12:24: hugeParam: metrics is heavy (96 bytes); consider passing it by pointer (gocritic)
```

## Dependencies

n/a

## Checklist

### DIFF
--- a/pkg/flow/account.go
+++ b/pkg/flow/account.go
@@ -100,7 +100,7 @@ func (c *Accounter) evict(entries map[ebpf.BpfFlowId]*ebpf.BpfFlowMetrics, evict
 	monotonicNow := uint64(c.monoClock())
 	records := make([]*model.Record, 0, len(entries))
 	for key, metrics := range entries {
-		flowContent := model.NewBpfFlowContent(*metrics)
+		flowContent := model.NewBpfFlowContent(metrics)
 		records = append(records, model.NewRecord(key, &flowContent, now, monotonicNow, c.s))
 	}
 	c.metrics.EvictionCounter.WithSourceAndReason("accounter", reason).Inc()
--- a/pkg/model/flow_content.go
+++ b/pkg/model/flow_content.go
@@ -9,8 +9,8 @@ type BpfFlowContent struct {
 	AdditionalMetrics *ebpf.BpfAdditionalMetrics
 }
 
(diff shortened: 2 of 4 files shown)


jotak:
@msherif1234 yeah I saw that linter warning but didn't change it ... at the end of the day, we still need to copy the object, which I guess is similar than passing by value. So we could as well make a //nolint here?

msherif1234:
> @msherif1234 yeah I saw that linter warning but didn't change it ... at the end of the day, we still need to copy the object, which I guess is similar than passing by value. So we could as well make a //nolint here?

I don't think its a good idea to ignore this lint error specialy this structure will keep growing and we should pass by ref instead of value

jotak:
> > @msherif1234 yeah I saw that linter warning but didn't change it ... at the end of the day, we still need to copy the object, which I guess is similar than passing by value. So we could as well make a //nolint here?
> 
> I don't think its a good idea to ignore this lint error specialy this structure will keep growing and we should pass by ref instead of value

My point is that making an explicit copy just before passing by ref looks pretty much similar to passing by value

msherif1234:
> ```shell
> hugeParam: metrics is heavy (96 bytes); consider passing it by pointer (gocritic)
> ```

passing by value will generate the copy from the whole structure I agree its the same but if we do it we keep code clean and don't have to skip lint there might better reasons too I don't know

jotak:
do as you wish, I've set the lgtm
Going with a pointer adds an indirection, and the fact that you need to copy anyway removes all the optimization benefit.
Also, the 80 bytes limit seems quite debated on the web (https://www.reddit.com/r/golang/comments/19f8ngi/gocritic_hugeparam_default_of_80_useful/ or https://github.com/go-critic/go-critic/issues/1200 )

msherif1234:
> do as you wish, I've set the lgtm Going with a pointer adds an indirection, and the fact that you need to copy anyway removes all the optimization benefit. Also, the 80 bytes limit seems quite debated on the web (https://www.reddit.com/r/golang/comments/19f8ngi/gocritic_hugeparam_default_of_80_useful/ or [go-critic/go-critic#1200](https://github.com/go-critic/go-critic/issues/1200) )
its more of nit just would like to keep code linter clean maybe its fine to disable linter for this instance anyway not urgent I will close this PR and might just disable the linter err for this function in a following PR

PR #538, opened 2025-01-30
NETOBSERV-2198: IPsec support

msherif1234:
## Description

probing XFRM input and output processing to know if packets came encrypted or left encrypted

image

image

```sh
[root@ip-10-0-1-35 /]# ip xfrm state
src 10.0.1.158 dst 10.0.1.35
	proto esp spi 0x17e83b29 reqid 16393 mode transport
	replay-window 0 flag esn
	aead rfc4106(gcm(aes)) 0x2dff2e25143878e71710cc484f87a0e04a5d9882b9e23643f28692e33935eef6879f7ef5 128
	anti-replay esn context:
	 seq-hi 0x0, seq 0x0, oseq-hi 0x0, oseq 0x0
	 replay_window 128, bitmap-length 4
	 00000000 00000000 00000000 00000000 
	sel src 10.0.1.158/32 dst 10.0.1.35/32 proto udp sport 6081 
```
## Dependencies

n/a

## Checklist


pkg/decode/decode_protobuf.go
```diff
@@ -141,6 +141,10 @@ func RecordToMap(fr *model.Record) config.GenericMap {
 			out["XlatSrcAddr"] = model.IP(fr.Metrics.AdditionalMetrics.TranslatedFlow.Saddr).String()
 			out["XlatDstAddr"] = model.IP(fr.Metrics.AdditionalMetrics.TranslatedFlow.Daddr).String()
 		}
+		if fr.Metrics.AdditionalMetrics.FlowEncrypted || fr.Metrics.AdditionalMetrics.FlowEncryptedRet != 0 {
+			out["EncryptedFlow"] = fr.Metrics.AdditionalMetrics.FlowEncrypted
+			out["EncryptedFlowRet"] = fr.Metrics.AdditionalMetrics.FlowEncryptedRet
```
jpinsonneau, line 146: I'm not a huge fan of the field naming here. Since the feature is called `IPSec` why not using `IPSecEncrypted` (or just `IPSec`) and `IPSecCode` ?

bpf/ipsec.h
```diff
@@ -0,0 +1,186 @@
...
+    BPF_PRINTK("Enter xfm dir: %d protocol: %d family: %d if_index: %d \n", dir, protocol, family,
+               if_index);
+
+    if (dir == INGRESS) {
+        ret = bpf_map_update_elem(&ipsec_ingress_map, &pid_tgid, &id, BPF_NOEXIST);
+    } else {
+        ret = bpf_map_update_elem(&ipsec_egress_map, &pid_tgid, &id, BPF_NOEXIST);
+    }
+    if (ret != 0) {
+        if (trace_messages && ret != -EEXIST) {
```
jotak, line 137: is there a reason for having `ret != -EEXIST` here? I guess it shouldn't happen? (If it does happen, then maybe we should use `BPF_ANY` in the `bpf_map_update_elem` calls above?)

bpf/ipsec.h
```diff
@@ -0,0 +1,186 @@
...
+        fill_in_others_protocol(&id, protocol);
+    }
+
+    // check if this packet need to be filtered if filtering feature is enabled
+    bool skip = check_and_do_flow_filtering(&id, flags, 0, eth_protocol, NULL, dir);
+    if (skip) {
+        return 0;
+    }
+
+    BPF_PRINTK("Enter xfm dir: %d protocol: %d family: %d if_index: %d \n", dir, protocol, family,
```
jotak, line 128: nit:
```suggestion
    BPF_PRINTK("Enter xfrm dir: %d protocol: %d family: %d if_index: %d \n", dir, protocol, family,
```

bpf/ipsec.h
```diff
@@ -0,0 +1,186 @@
...
+#include "utils.h"
+
+static inline int ipsec_lookup_and_update_flow(flow_id *id, int flow_encrypted_ret,
+                                               u16 eth_protocol) {
+    additional_metrics *extra_metrics = bpf_map_lookup_elem(&additional_flow_metrics, id);
+    if (extra_metrics != NULL) {
+        extra_metrics->end_mono_time_ts = bpf_ktime_get_ns();
+        extra_metrics->eth_protocol = eth_protocol;
+        extra_metrics->flow_encrypted_ret = flow_encrypted_ret;
+        extra_metrics->flow_encrypted = flow_encrypted_ret == 0 ? true : false;
```
jotak, line 17: if a flow contains both encrypted and unencrypted packets, here we keep only the last packet state.
I'd argue it's more important to show any unencrypted state, no?
So something like:
```
if (extra_metrics->flow_encrypted && flow_encrypted_ret != 0) {
extra_metrics->flow_encrypted = false;
}
```
wdyt?

bpf/types.h
```diff
@@ -140,6 +140,8 @@ typedef struct additional_metrics_t {
     } translated_flow;
     u16 eth_protocol;
     u8 network_events_idx;
+    bool flow_encrypted;
+    u8 flow_encrypted_ret;
```
jotak, line 144: if flow_encrypted is entirely derived from flow_encrypted_ret, maybe we can only keep the ret, and leave the interpretation to the user space? (as you prefer, I'm fine either way)

bpf/maps_definition.h
```diff
@@ -79,4 +79,24 @@ struct {
     __uint(pinning, LIBBPF_PIN_BY_NAME);
 } peer_filter_map SEC(".maps");
 
+// HashMap to store ingress flowid to be able to retrieve them from kretprobe hook
+struct {
+    __uint(type, BPF_MAP_TYPE_HASH);
+    __uint(max_entries, 1 << 20); // Will take around 64MB of space.
```
jotak, line 85: if I'm correct, the map content only lives during the xfrm call .... so maybe 64MB is over-sized? wdyt?

bpf/types.h
jotak: hmmm I guess there's the extra case where IPSec isn't involved so flow_encrypted==false and flow_encrypted_ret==0 which we need to account for, we couldn't do that with just flow_encrypted_ret. OK so forget about my previous comment :-)

bpf/ipsec.h
msherif1234: I don't think I should see EEXIST error I will remove this check

bpf/ipsec.h
msherif1234: the encrypted flag name my be missleading here as it work both ways
for xfrm_input where we decrypt and check decryption state, while in xfrm_out this the encryption path , maybe I should rename them here as well to ipsec_ret and ipsec_success similar to the json ?

bpf/maps_definition.h
msherif1234: its NO_PREALLOC map types i,e allocate as needed so there should be no issues with size right ?

pkg/model/flow_content.go
```diff
@@ -117,6 +117,13 @@ func (p *BpfFlowContent) AccumulateAdditional(other *ebpf.BpfAdditionalMetrics)
 	if !AllZeroIP(IP(other.TranslatedFlow.Saddr)) && !AllZeroIP(IP(other.TranslatedFlow.Daddr)) {
 		p.AdditionalMetrics.TranslatedFlow = other.TranslatedFlow
 	}
+	// Encryption
+	if other.FlowEncrypted {
+		p.AdditionalMetrics.FlowEncrypted = other.FlowEncrypted
+	}
+	if p.AdditionalMetrics.FlowEncryptedRet != other.FlowEncryptedRet {
+		p.AdditionalMetrics.FlowEncryptedRet = other.FlowEncryptedRet
+	}
```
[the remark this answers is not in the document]
msherif1234: won't the above show both ?

bpf/types.h
msherif1234: sorry I am confused if both fields are default this is the no ipsec case right ?

bpf/maps_definition.h
jotak: ok

bpf/types.h
jotak: yes - no problem here

jotak (approved):
/lgtm

PR #552, opened 2025-02-07
Use CAS to set do_sampling for automicty

msherif1234:
## Description

setting up do_sampling isn't atomic this PR uses CAS operation to set it
## Dependencies

n/a

## Checklist


jotak:
I'm not sure to understand how CAS would solve the problem that we are facing. The problem is that multiple CPU are updating the same variable, and `do_sampling` only hold 1 state whereas there should be 1 state per CPU (or per packet, I'm not sure) to keep. So we're still trying to fit multiple sampling decisions into a single variable

jotak:
I'm wondering if setting `do_sampling` in a per-cpu map (a map of just 1 value) could work ... depends if the same CPU is used for the same packet processing, I'm not sure about that.

msherif1234:
> I'm wondering if setting `do_sampling` in a per-cpu map (a map of just 1 value) could work ... depends if the same CPU is used for the same packet processing, I'm not sure about that.
today `do_sampling` is more of field in global map anyway we need to think more about this we probably should track this with Jira card and I can take it

msherif1234:
> I'm not sure to understand how CAS would solve the problem that we are facing. The problem is that multiple CPU are updating the same variable, and `do_sampling` only hold 1 state whereas there should be 1 state per CPU (or per packet, I'm not sure) to keep. So we're still trying to fit multiple sampling decisions into a single variable

yeah the flow state is global now hence the do_sampling the CAS was just to make it thread safe like how we update the flow map value today

msherif1234:
not needed

PR #587, opened 2025-02-26
NETOBSERV-2092 Inject drop events in drop fields

jpinsonneau:
## Description

Force OVS_DROP cause when received a drop event.
Also set dropped bytes and packets to have proper metrics.

Original discussion: https://github.com/netobserv/netobserv.github.io/pull/6#discussion_r1939152634

## Dependencies

n/a

## Checklist

### DIFF
--- a/pkg/decode/decode_protobuf.go
+++ b/pkg/decode/decode_protobuf.go
@@ -149,6 +149,14 @@ func RecordToMap(fr *model.Record) config.GenericMap {
 
 	if len(fr.NetworkMonitorEventsMD) != 0 {
 		out["NetworkEvents"] = fr.NetworkMonitorEventsMD
+		for _, event := range fr.NetworkMonitorEventsMD {
+			// override drop fields when network event action is dropped
+			if event["Action"] == "drop" {
+				out["PktDropBytes"] = fr.Metrics.Bytes
+				out["PktDropPackets"] = fr.Metrics.Packets
+				out["PktDropLatestDropCause"] = "OVS_DROP_EXPLICIT"
+			}
+		}
 	}
 
 	return out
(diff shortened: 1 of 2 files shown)


pkg/decode/decode_protobuf.go
```diff
@@ -149,6 +149,14 @@ func RecordToMap(fr *model.Record) config.GenericMap {
 
 	if len(fr.NetworkMonitorEventsMD) != 0 {
 		out["NetworkEvents"] = fr.NetworkMonitorEventsMD
+		for _, event := range fr.NetworkMonitorEventsMD {
+			// override drop fields when network event action is dropped
+			if event["Action"] == "drop" {
+				out["PktDropBytes"] = fr.Metrics.Bytes
```
msherif1234, line 155: This is not right the same flow could be going through allow ingress and drop egress events ?

jpinsonneau: Could we have the amount of packets / bytes from the event ?

Previously we had these amounts of drops. This is displayed in the plugin + dashboards so it would be very useful to get that back somehow

msherif1234: its not a single event same flow can have allow and deny event or 2 allow and 1 deny or any other combo so hard to tell, the drop we had will still work for everything else except netpol and adminnetpol right ?

jpinsonneau: Yes of course but the goal here is to be able to make graphs with the amount of bytes / packets / flows denied because of policies.

jpinsonneau: That's how @jotak managed it for now: https://github.com/netobserv/network-observability-operator/pull/863 + https://github.com/netobserv/flowlogs-pipeline/pull/739

The console plugin could rely on prometheus for these but will not be able to query loki without extra processing

jpinsonneau:
Closing this but keeping the jira as we still need a solution here

PR #616, opened 2025-03-18
nit: unused Duplicate field in protobuf

jotak:
This is just to clarify in code comments that it's now unused


jotak:
merging, it's really just a code comment

Issue #617, opened 2025-03-18
flp pods are going crashloopbackoff


ajayanaj:
Hello,

After k8's version 1.31 update we noticed that our flp pods are going to crashloopbackoff state. Upon checking found the below error.

```
time=2025-03-18T15:54:35Z level=debug msg=opts.PipeLine = [{"name":"ingest"},{"follows":"ingest","name":"enrich"},{"follows":"enrich","name":"loki"}]
time=2025-03-18T15:54:35Z level=debug msg=stages = [{ingest } {enrich ingest} {loki enrich}]
time=2025-03-18T15:54:35Z level=error msg=error when parsing pipeline parameters: json: unknown field "input"
time=2025-03-18T15:54:35Z level=error msg=error in parsing config file: json: unknown field "input"
```

We used the same configmap as mentioned here https://github.com/netobserv/netobserv-ebpf-agent/blob/main/deployments/flp-daemonset-cap.yml#L77-L121
We use latest image: quay.io/netobserv/flowlogs-pipeline:main
I couldn't figure out what's wrong with this. 
Anyone please help me here?

jotak:
Hi @ajayanaj , it's not recommended to use `quay.io/netobserv/flowlogs-pipeline:main` as there could be breaking change any time. You can try with an older version such as `quay.io/netobserv/flowlogs-pipeline:v1.8.1-community ` or `quay.io/netobserv/flowlogs-pipeline:v1.8.0-community` or `quay.io/netobserv/flowlogs-pipeline:v1.8.1-community ` or `quay.io/netobserv/flowlogs-pipeline:v1.7.0-community`

I'm looking at the latest breaking changes to see what it could be ...

jotak:
I think it must be that: https://github.com/netobserv/flowlogs-pipeline/pull/711
Previously, in the k8s configuration, you had probably something like that: 
```
    kubernetes:
      input:  "SrcAddr",
```

"input" must be changed to "ipField":
```
    kubernetes:
      ipField:  "SrcAddr",
```

jotak:
by the way, you're right that the example config here is outdated, it should be changed as well

ajayanaj:
Yes, input was the issue after changing the same to "ipField" it started working. 
Does this image works with "ipField" quay.io/netobserv/flowlogs-pipeline:v1.8.1-community ?

jotak:
yes

ajayanaj:
Thanks @jotak

PR #618, opened 2025-03-19
Fix outdated examples

jotak:
See also: #617


jotak:
Trivial changes / no code

PR #682, opened 2025-04-28
NETOBSERV-2193: fix race on capacity limiter test

jotak:
The race was, some elements went out of the pipe before everything was in, allowing more than capacity being processed

### DIFF
--- a/pkg/flow/limiter_test.go
+++ b/pkg/flow/limiter_test.go
@@ -2,6 +2,7 @@ package flow
 
 import (
 	"strconv"
+	"sync"
 	"testing"
 
 	"github.com/netobserv/gopipes/pkg/node"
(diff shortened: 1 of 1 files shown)


pkg/flow/limiter_test.go
```diff
@@ -59,23 +60,48 @@ func TestCapacityLimiter_Drop(t *testing.T) {
...
 	}
 }
 
-func capacityLimiterPipe() (in chan<- []*model.Record, out <-chan []*model.Record) {
+func capacityLimiterPipe() (chan<- []*model.Record, <-chan []*model.Record) {
 	inCh, outCh := make(chan []*model.Record), make(chan []*model.Record)
 
 	init := node.AsInit(func(initOut chan<- []*model.Record) {
 		for i := range inCh {
+			// fmt.Printf("in: %s\n", i[0].Interfaces[0].Interface)
```
jpinsonneau, line 78:
```suggestion
```

pkg/flow/limiter_test.go
```diff
@@ -59,23 +60,48 @@ func TestCapacityLimiter_Drop(t *testing.T) {
...
+				}
+				prev = n
+				time.Sleep(50 * time.Millisecond)
+			}
+		}()
+		wg.Wait()
+
+		// Start output
 		for i := range termIn {
+			// fmt.Printf("out: %s\n", i[0].Interfaces[0].Interface)
```
jpinsonneau, line 104:
```suggestion
```

jpinsonneau (approved):
LGTM, just remove the comented code before merging

pkg/flow/limiter_test.go
jotak: I wanted to keep it because I'm not 100% sure it's a definitive fix (it's always the problem with race conditions...) and in case I see it again, it'll be faster to debug again

jotak: anyway, ok I can remove them, hopefully the fix is definitive :-)

PR #683, opened 2025-04-28
Disable cgo on upstream builds and attach binaries to releases

jotak:
See also (same changes): https://github.com/netobserv/flowlogs-pipeline/pull/932


msherif1234:
/lgtm
I can't see why konflux is failing pls check the log to confirm its not related to the changes

jotak:
konflux seems quite unstable today, I can't see the logs because they disappear super quickly, IMO that looks like worker provisioning issues... not related to this PR

PR #693, opened 2025-05-06
NETOBSERV-2101: document needed caps, remove CAP_SYS_RESOURCE

jotak:
Hey @msherif1234 , FYI I picked up your descriptions from [there](https://github.com/netobserv/network-observability-operator/blob/main/controllers/ebpf/internal/permissions/permissions.go#L21-L24), but also, while testing I'm finding that SYS_RESOURCE isn't needed anymore. Not sure if that's a moving requirement related to a code change from us, or from the linux kernel, or something else ...

Anyway, I'll the change also in the operator and we'll verify that with tests on different ocp versions and enabled features.

See also: https://github.com/netobserv/network-observability-operator/pull/1484

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,9 +82,12 @@ To deploy it as a Pod, you can check the [deployment examples](./deployments).
 
 The Agent needs to be executed either with:
 
-1. The following [Linux capabilities](https://man7.org/linux/man-pages/man7/capabilities.7.html)
-   (recommended way): `BPF`, `PERFMON`, `NET_ADMIN`, `SYS_RESOURCE`. If you
-   [deploy it in Kubernetes or OpenShift](./deployments/flp-daemonset-cap.yml),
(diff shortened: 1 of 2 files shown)


msherif1234:
which kernel did u test with I assume rhel9.6 ? when I tested with  rhel9.4 which is what we use for 1.8 and OCP4.18 I wasn't able to drop RESOURCE_CAP can run ur pr with 4.18 and see if that is still the case ?

jotak:
I used a 4.18 cluster. Do you remember if it was a particular feature that was failing? 
We definitely need to run broad tests before merging it

OlivierCazade (approved):
LGTM

memodi:
/label qe-approved

verified on older kernel 4.12 and 4.14 OCP versions, all sanity tests passes.

jotak:
thanks!

PR #696, opened 2025-05-12
Fix pca - need to unpin ipsec maps

jotak:
Otherwise we're getting this error while running pca:
```
time="2025-05-12T11:03:12Z" level=fatal msg="[PCA] can't instantiate NetObserv eBPF Agent" error="loading and assigning BPF objects: field IpsecEgressMap: map ipsec_egress_map: pin by name: missing MapOptions.PinPath"
```

### DIFF
--- a/pkg/tracer/tracer.go
+++ b/pkg/tracer/tracer.go
@@ -1501,6 +1501,10 @@ func NewPacketFetcher(cfg *FlowFetcherConfig) (*PacketFetcher, error) {
 			TcpRcvKprobe:            nil,
 			KfreeSkb:                nil,
 			NetworkEventsMonitoring: nil,
+			XfrmInputKretprobe:      nil,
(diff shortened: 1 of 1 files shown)


jpinsonneau (approved):
Thanks @jotak !

msherif1234:
oh sorry my mistake while u there pls init the hook to nil as well in https://github.com/netobserv/netobserv-ebpf-agent/blob/main/pkg/tracer/tracer.go#L1485

jotak:
@msherif1234 like this: [c9ac6f6](https://github.com/netobserv/netobserv-ebpf-agent/pull/696/commits/c9ac6f605e06b3cd59e0cf1b7dda65191fd6ab33) ? I guess that's not necessary (they should be 0-valued as nil anyway, right?) anyway, sounds better for consistency

PR #703, opened 2025-05-19
NETOBSERV-2257: identify interfaces by index and mac

jotak:
## Description

Index alone is not safe to identify interfaces, as they are not unique across multiple netns.

In our tests with ovn-k, we've seen in particular eth0 and ens5 regularly colliding, both using ifindex 2. This leads to inconsistencies e.g. when filtering with exlucded interface ens5, there was still traffic (wrongly) identified as coming from ens5 (it was actually eth0).

To fix it, when a collision is seen, interface naming uses MAC address to disambiguate.

## Dependencies

operator: https://github.com/netobserv/network-observability-operator/pull/1565

## Checklist

  * [x] If so, fill in "Release Note Text" in the JIRA.


pkg/ifaces/informer.go
```diff
@@ -38,6 +39,7 @@ type Event struct {
 type Interface struct {
 	Name   string
 	Index  int
+	MAC    [6]uint8
```
msherif1234, line 42: @jotak last time I added field there nsString performance degraded, can pls check perf impact with this change to be sure ?

msherif1234:
nit: what happens if u kept mac type as `net.HardwareAddr` instead of converting it to [6] ?

jotak:
> nit: what happens if u kept mac type as `net.HardwareAddr` instead of converting it to [6] ?

I use it in a map for indexing I can't use a slice as index but fixed-size array is ok

pkg/ifaces/informer.go
jotak: sure, @memodi triggered a test but it failed, let me retry...

jotak: ouch, bad news, there's +77% CPU in DNH :-/
... going to investigate that ... I'm surprised because doing just 2 maps lookup in the worst case didn't seem bad to me ; or maybe that's the fallback code invoked more than I thought

jotak: I've found the problem I think. Problem is with OVN as it makes up MAC based on IPs in eth header, even when the if_index is 2 (eth0). So the flows MAC don't correspond in that case to the eth0 interface MAC, which triggers a cache miss, and fallback to syscall, which is why the CPU increases.
I've added an OVN specific optimization to avoid these syscalls in that situation, see last commit

pkg/ifaces/registerer.go
```diff
@@ -55,19 +71,54 @@ func (r *Registerer) Subscribe(ctx context.Context) (<-chan Event, error) {
...
+		} else if len(macsMap) > 1 {
+			// Several entries, need to disambiguate by MAC
+			name, ok := macsMap[mac]
+			if ok {
+				return name, true
+			}
+			// Not found => before falling back to syscall lookup that is CPU intensive, run this quick ovn optimization:
+			// eth0 & ens5 often collide; MAC starting with 0A:58 should be eth0 and not ens5, but since the MAC captured in flow
+			// doesn't match the actual interface MAC, we'll hardcode that.
+			if mac[0] == 0x0a && mac[1] == 0x58 {
```
msherif1234, line 93: I don't think this an endian neutral way comparing mac this way ?

jotak: I'm not super knowledgeable on endianess by I would say endianess doesn't affect byte arrays? It matters for bigger-than-byte types, like ints, but with a byte array _we_ control what's on the lower and upper index, this is not left to the architecture I think? E.g. an int16 `0x1234` could be `[0x12, 0x34]` or `[0x34, 0x12]` but a byte array `[0x12, 0x34]` always stays as is? I guess there could be an issue if macs were provided as ints but afaik it's never the case.
Again, I'm not super confident on this topic ...

msherif1234: > case.

Just take note of this to be verified on s390x as it's the only BE arch we have atm

msherif1234, line 93: one last comment in this PR since we don't know how many more exceptions will be needed may I suggest have a slice of those hardcode value so new exceptions will be just adding to teh slice w/o any code changes ?

jotak: you mean, to have it configurable instead of hardcoded?

jotak: I made it configurable - so now needs also this operator PR https://github.com/netobserv/network-observability-operator/pull/1565

jotak:
merging before branch cut

PR #716, opened 2025-06-05
Prevent lost updates in IfaceNameForIndexAndMAC

frobware:
This change addresses a concurrency bug in IfaceNameForIndexAndMAC where multiple goroutines could simultaneously observe a missing cache entry, perform a fallback syscall, and then overwrite each other’s results. The symptom is not a data race but a logical “lost update”, where one thread clobbers the data written by another.

To prevent this, the function now uses double-checked locking: after performing the fallback lookup, it re-checks the cache under a write lock before inserting.

No unit test has been added - while the bug is real, it’s difficult to trigger in practice and cannot be reliably reproduced without artificial coordination between goroutines. The behaviour is safe in terms of memory access (thanks to proper locking), but was incorrect in terms of state consistency.

This fix ensures correctness under concurrent access without sacrificing performance.

### DIFF
--- a/pkg/ifaces/registerer_test.go
+++ b/pkg/ifaces/registerer_test.go
@@ -143,7 +143,8 @@ func TestRegisterer_Lookup(t *testing.T) {
 	assert.False(t, ok)
 }
 
-func TestRace_Registerer_Lookup(t *testing.T) {
+func TestRegisterer_LookupRace(t *testing.T) {
+	// No assertion here, purpose being to catch races
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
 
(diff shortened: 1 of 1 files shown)


pkg/ifaces/registerer.go
```diff
@@ -78,9 +78,37 @@ func (r *Registerer) Subscribe(ctx context.Context) (<-chan Event, error) {
...
+// Concurrency note:
+//
+// Without double-checked locking, the following sequence may occur:
+//
+//  1. Goroutine A acquires RLock, sees r.ifaces[idx][mac] is missing
+//  2. Goroutine B does the same (also sees entry missing)
+//  3. Both release RLock and call net.InterfaceByIndex(idx)
+//  4. Both prepare to insert iface.Name into r.ifaces[idx][mac]
+//  5. Goroutine A acquires Lock and writes to the map
+//  6. Goroutine B acquires Lock and overwrites A's result
```
jotak, line 107: I'm not sure?
If both goroutines have the same `idx`,
- if they have same MAC, yes B would overwrite A, but with the same value, so that's not an issue / there's no loss of information
- if they have different MAC, there's no overwrite

?

frobware: So the question becomes: is my description / sequencing described badly but the issue persists. Or, it's always the same value so in reality it doesn't matter? It does seem like two goroutines can get to the same point and then one wins by overwriting.

PR #730, opened 2025-06-23
Revert grpc bump

jotak:
grpc 1.73.0 seems to degrade memory usage (to be confirmed)

See also: https://github.com/grpc/grpc-go/issues/8403


jotak:
although it's not clear if this was causing a regression, I'll go ahead and merge this for release-1.9, as there wasn't any real need to bump here.

PR #815, opened 2025-10-14
NETOBSERV-2471: TLS usage tracking

jotak:
- Start implementing TLS, by reading the TLS header when present
- Extract SSL version
- Report the TLS version in output records

Fixes https://github.com/netobserv/network-observability-operator/issues/2492

## Dependencies

- Operator: https://github.com/netobserv/network-observability-operator/pull/2124
- FLP: https://github.com/netobserv/flowlogs-pipeline/pull/1119
- Console: https://github.com/netobserv/netobserv-web-console/pull/1393

## Summary by CodeRabbit

* **New Features**
  * TLS/SSL usage tracking added (config option ENABLE_TLS_TRACKING, default off).
  * Flow records now include TLS metadata: version, cipher suite, key-exchange/curve, TLS message types, and SSL-mismatch indication.
  * Export/protobuf and decoding now surface TLS fields so downstream consumers and UIs can display TLS info.

* **Chores**
  * Updated vendored dependencies including OpenTelemetry and various golang modules.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -162,5 +162,3 @@ require (
 )
 
 replace github.com/vmware/go-ipfix => github.com/jotak/go-ipfix v0.0.0-20250708115123-407c539ea101
-
-replace github.com/netobserv/flowlogs-pipeline => github.com/jotak/flowlogs-pipeline v0.0.0-20260319155004-bff214fed5a0
--- a/go.sum
+++ b/go.sum
@@ -177,8 +177,6 @@ github.com/josharian/intern v1.0.0/go.mod h1:5DoeVV0s6jJacbCEi61lwdGj/aVlrQvzHFF
 github.com/josharian/native v1.0.0/go.mod h1:7X/raswPFr05uY3HiLlYeyQntB6OO7E/d2Cu7qoaN2w=
 github.com/josharian/native v1.1.0 h1:uuaP0hAbW7Y4l0ZRQ6C9zfb7Mg1mbFKry/xzDAfmtLA=
 github.com/josharian/native v1.1.0/go.mod h1:7X/raswPFr05uY3HiLlYeyQntB6OO7E/d2Cu7qoaN2w=
-github.com/jotak/flowlogs-pipeline v0.0.0-20260319155004-bff214fed5a0 h1:KRvLD3FlAK68qyDfC/AI/LvNcYmJf4ZppTeTOlRJhM0=
-github.com/jotak/flowlogs-pipeline v0.0.0-20260319155004-bff214fed5a0/go.mod h1:XSjKtr+AVl8/owIidOPkO97KD9/ygNByuUddFaiXmb8=
 github.com/jotak/go-ipfix v0.0.0-20250708115123-407c539ea101 h1:tpaHjydMAy2MTukKIUAVK4xIFUpL12xuexA0FuTVpuo=
 github.com/jotak/go-ipfix v0.0.0-20250708115123-407c539ea101/go.mod h1:GgFbcmEGqMQfA7jDC9UVLKAelNh2sy1jsxyV7Tor3Ig=
 github.com/jpillora/backoff v1.0.0 h1:uvFg412JmmHBHw7iwprIxkPMI+sGQ4kzOWsMeHnm2EA=
@@ -261,6 +259,8 @@ github.com/mwitkow/go-conntrack v0.0.0-20190716064945-2f068394615f h1:KUppIJq7/+
 github.com/mwitkow/go-conntrack v0.0.0-20190716064945-2f068394615f/go.mod h1:qRWi+5nqEBWmkhHvq77mSJWrCKwh8bxhgT7d/eI7P4U=
 github.com/mxk/go-flowrate v0.0.0-20140419014527-cca7078d478f h1:y5//uYreIhSUg3J1GEMiLbxo1LJaP8RfCpH6pymGZus=
 github.com/mxk/go-flowrate v0.0.0-20140419014527-cca7078d478f/go.mod h1:ZdcZmHo+o7JKHSa8/e818NopupXU1YMK5fe1lsApnBw=
+github.com/netobserv/flowlogs-pipeline v1.11.2-community h1:OdcKKyvfTWtoFIQk2tQEWlnfEuLdJTmskffM/RKmlMk=
+github.com/netobserv/flowlogs-pipeline v1.11.2-community/go.mod h1:j6eOnbQBcUkWqmP20BNSUuQE/UQMEL146yYVHHtmECQ=
 github.com/netobserv/gopipes v0.3.0 h1:IYmPnnAVCdSK7VmHmpFhrVBOEm45qpgbZmJz1sSW+60=
 github.com/netobserv/gopipes v0.3.0/go.mod h1:N7/Gz05EOF0CQQSKWsv3eof22Cj2PB08Pbttw98YFYU=
 github.com/netobserv/loki-client-go v0.0.0-20251014110557-40bc8d2e6cf3 h1:rxQipq0xpoiao7ifls/82JCcOVALC4n08ppTLCUFGL4=
--- a/vendor/modules.txt
+++ b/vendor/modules.txt
@@ -294,7 +294,7 @@ github.com/mwitkow/go-conntrack
 # github.com/mxk/go-flowrate v0.0.0-20140419014527-cca7078d478f
 ## explicit
 github.com/mxk/go-flowrate/flowrate
-# github.com/netobserv/flowlogs-pipeline v1.11.2-community => github.com/jotak/flowlogs-pipeline v0.0.0-20260319155004-bff214fed5a0
+# github.com/netobserv/flowlogs-pipeline v1.11.2-community
 ## explicit; go 1.25.0
 github.com/netobserv/flowlogs-pipeline/pkg/api
 github.com/netobserv/flowlogs-pipeline/pkg/config
@@ -1396,4 +1396,3 @@ sigs.k8s.io/structured-merge-diff/v6/value
 sigs.k8s.io/yaml
 sigs.k8s.io/yaml/goyaml.v2
 # github.com/vmware/go-ipfix => github.com/jotak/go-ipfix v0.0.0-20250708115123-407c539ea101
-# github.com/netobserv/flowlogs-pipeline => github.com/jotak/flowlogs-pipeline v0.0.0-20260319155004-bff214fed5a0


bpf/utils.h
```diff
@@ -50,6 +50,22 @@ static inline void set_flags(struct tcphdr *th, u16 *flags) {
...
+    void *start_payload = ((void *)tcp) + (tcp->doff * 4);
+    if (start_payload + 5 <= data_end) {
+        if (((u8 *)start_payload)[0] == 0x16) {
+            // TODO handshake special case, see https://www.netmeister.org/blog/tcpdump-ssl-and-tls.html
+            pkt->ssl_version =
+                ((u16)(((u8 *)start_payload)[1])) << 8 | (u16)(((u8 *)start_payload)[2]);
+        } else if (((u8 *)start_payload)[0] == 0x14 || ((u8 *)start_payload)[0] == 0x15 ||
+                   ((u8 *)start_payload)[0] == 0x17) {
+            pkt->ssl_version =
+                ((u16)(((u8 *)start_payload)[1])) << 8 | (u16)(((u8 *)start_payload)[2]);
```
jotak: (need to check endianness for that)

bpf/flows.c
```diff
@@ -178,6 +194,7 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
     if (enable_dns_tracking) {
         dns_errno = track_dns_packet(skb, &pkt);
     }
+    track_tls_version(skb, &pkt);
```
msherif1234: don't think its better if we have feature config for tracker like we do with all other features ? instead of enabling it by default  ?

bpf/flows.c
```diff
@@ -93,6 +98,17 @@ static __always_inline void update_existing_flow(flow_metrics *aggregate_flow, p
...
         aggregate_flow->sampling = sampling;
+        if (pkt->ssl_version != 0 && aggregate_flow->ssl_version != pkt->ssl_version) {
+            if (aggregate_flow->ssl_version == 0) {
+                aggregate_flow->ssl_version = pkt->ssl_version;
+            } else {
+                // If several SSL versions are found, keep just the smallest and set the "mismatch" flag
+                if (pkt->ssl_version < aggregate_flow->ssl_version) {
+                    aggregate_flow->ssl_version = pkt->ssl_version;
+                }
+                aggregate_flow->misc_flags |= MISC_FLAGS_SSL_MISMATCH;
```
msherif1234, line 111: is it better to have mismatch counter instead of flag ?

bpf/tls_tracker.h
```diff
@@ -0,0 +1,82 @@
...
+};
+
+struct tls_handshake_version {
+    u8 major;
+    u8 minor;
+};
+
+// Extract TLS info
+static inline void track_tls_version(struct __sk_buff *skb, pkt_info *pkt) {
+    if (pkt->id->transport_protocol == IPPROTO_TCP) {
```
msherif1234: do u think this work for `STCP` protocol too ?, not sure what is the story for DTLS ?

bpf/tls_tracker.h
```diff
@@ -0,0 +1,82 @@
...
+        void *data_end = (void *)(long)skb->data_end;
+        struct tcphdr *tcp = (struct tcphdr *)pkt->l4_hdr;
+        if (!tcp || ((void *)tcp + sizeof(*tcp) > data_end)) {
+            return;
+        }
```
msherif1234, line 61: does the verifier needs the above check given how late we call this tracker all headers should have been sane already ?

bpf/tls_tracker.h
```diff
@@ -0,0 +1,82 @@
...
+    u8 len[3];
+};
+
+struct tls_handshake_version {
+    u8 major;
+    u8 minor;
+};
+
+// Extract TLS info
+static inline void track_tls_version(struct __sk_buff *skb, pkt_info *pkt) {
```
msherif1234: might be good idea to add return code and track failure at the caller since there many conditions check in this function and if things doesn't work it will be hard to debug.
might be good adding BPF_PRINTK() on failing checks ?

bpf/tls_tracker.h
```diff
@@ -0,0 +1,82 @@
...
+        struct tls_record rec;
+        u32 offset = (long)pkt->l4_hdr - (long)skb->data + len;
+
+        if ((bpf_skb_load_bytes(skb, offset, &rec, sizeof(rec))) < 0) {
+            return;
+        }
+
+        switch (rec.content_type) {
+        case CONTENT_TYPE_HANDSHAKE: {
+            pkt->ssl_version = ((u16)rec.major) << 8 | rec.minor;
```
msherif1234: why setting it here if u really want handshake ssl version ?

bpf/tls_tracker.h
```diff
@@ -0,0 +1,82 @@
...
+                }
+                pkt->ssl_version = ((u16)handshake_version.major) << 8 | handshake_version.minor;
+            }
+            break;
+        }
+        case CONTENT_TYPE_CHANGE_CIPHER:
+        case CONTENT_TYPE_ALERT:
+        case CONTENT_TYPE_APP_DATA:
+            pkt->ssl_version = ((u16)rec.major) << 8 | rec.minor;
+            break;
```
msherif1234: do u think adding SSL record type to the follow will be adding value to end user ?

bpf/tls_tracker.h
```diff
@@ -0,0 +1,82 @@
...
+*/
+
+#ifndef __TLS_TRACKER_H__
+#define __TLS_TRACKER_H__
+#include "utils.h"
+
+#define CONTENT_TYPE_CHANGE_CIPHER 0x14
+#define CONTENT_TYPE_ALERT 0x15
+#define CONTENT_TYPE_HANDSHAKE 0x16
+#define CONTENT_TYPE_APP_DATA 0x17
```
msherif1234, line 16: based on this gist it seems there also heartbeat
https://gist.github.com/coin8086/1cd0411447066a5a02be6a3e493479e2

msherif1234: maybe something for the feature could be measuring the handshake latency ?

msherif1234 (commented):
Nice work!!, I left some comments/suggestions, nothing major small enhancements . Also pls share some screenshots for this new feature

bpf/tls_tracker.h
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        struct tls_record rec;
+        u32 offset = (long)pkt->l4_hdr - (long)skb->data + len;
+
+        if ((bpf_skb_load_bytes(skb, offset, &rec, sizeof(rec))) < 0) {
+            return;
+        }
+
+        switch (rec.content_type) {
```
msherif1234: This logic assume there is always tls record after tcp header this is not safe assumption we could have dns header here or any other header ??

bpf/flows.c
jotak: the flag allows us to correlate precisely with a particular flow, and report that up to the UI

bpf/flows.c
jotak: yes I'll do that (although I would like to have it enabled by default if it doesn't show visible impact on perfs)

bpf/tls_tracker.h
jotak: But that would require a new map like we do for DNS, it's probably more impact on performance? If so, it should probably be a separate feature

bpf/tls_tracker.h
jotak: I haven't digged into other protocols.. might be something to do if someone asks for it, but I believe TCP covers most of the expectations

bpf/tls_tracker.h
jotak: Yes I had a verifier error without those checks

bpf/tls_tracker.h
jotak: right, I should harden that a bit more; I've found unexpected TLSVersion during my tests probably because of that.

bpf/tls_tracker.h
jotak: that's a fallback for the case where handshake header couldn't be read ... But I'm going to change that, actually perhaps we don't want the client-hello version at all, it doesn't tell what's actually going to be used

bpf/tls_tracker.h
jotak: you mean adding it to the flow, like, as a bitfield? hmm why not ..

bpf/tls_tracker.h
```diff
@@ -0,0 +1,268 @@
...
+            if (ext_hdr.type == 0x002b) {
+                // Supported Versions: single version expected
+                u16 version;
+                if (bpf_skb_load_bytes(skb, offset + ext_offset, &version, sizeof(version)) < 0) {
+                    return TLSTRACKER_UNKNOWN;
+                }
+                tls->version = bpf_ntohs(version);
+                // Stop reading here
+                return TLSTRACKER_MATCHED;
+            }
```
rhmdnd, line 194: Something else that would be nice to pull out of the extensions is the key share. This would tell users if the server is using classical or hybrid key exchange algorithms, which is useful for finding out which applications are PQC-ready (e.g., looking for `X25519` + `Kyber768`).

We might need to do something like:

```c
if (ext_hdr.type == 0x002b) {
    // Supported Versions: single version expected
    u16 version;
    if (bpf_skb_load_bytes(skb, offset + ext_offset, &version, sizeof(version)) < 0) {
        return TLSTRACKER_UNKNOWN;
    }
    tls->version = bpf_ntohs(version);
} 
else if (ext_hdr.type == 0x0033) {
    // Key Share Extension (TLS 1.3)
    // The Server Hello Key Share contains:
    // Group (2 bytes) + Key Exchange Length (2 bytes) + Key Exchange Data
    u16 selected_group;
    if (bpf_skb_load_bytes(skb, offset + ext_offset, &selected_group, sizeof(selected_group)) < 0) {
        // not sure what would be best to return here?
        return KEYSHARE_UNKNOWN;
    }
    tls->key_share_group = bpf_ntohs(selected_group);
}
```

The `tls_info` would need to be updated to relay that information though.

Thoughts?

jotak: yep, definitely, if that's useful to figure out PQC-readiness, it's something that we want here
I'll add it, thanks for the hint!

jotak: done here [a9c6e61](https://github.com/netobserv/netobserv-ebpf-agent/pull/815/commits/a9c6e6160db3015b02b313a29091c06a223b3767)

jotak:
image

pkg/model/tls_types.go
```diff
@@ -0,0 +1,33 @@
...
+	{value: 32, name: "AppData"},
+}
+var typesMap map[string]uint8
+
+func init() {
+	typesMap = make(map[string]uint8, len(tlsTypes))
+	for _, t := range tlsTypes {
+		typesMap[t.name] = t.value
+	}
+}
```
jpinsonneau: Is that still used ?

bpf/tls_tracker.h
```diff
@@ -0,0 +1,285 @@
...
+        if (bpf_skb_load_bytes(skb, offset, &cipher_len, sizeof(cipher_len)) < 0) {
+            return TLSTRACKER_UNKNOWN;
+        }
+        offset += 2 + bpf_ntohs(cipher_len);
+        // Read compression
+        if (bpf_skb_load_bytes(skb, offset, &compr_len, sizeof(compr_len)) < 0) {
+            return TLSTRACKER_UNKNOWN;
+        }
+        offset += 1 + compr_len;
+        // Read extensions
```
jpinsonneau, line 73: It would be good to be able to see pre TLS-1.2 connections here even if it's not expected in recent environments. That's especially useful for auditing legacy TLS usage.

Something around those lines should work:

```suggestion
    u8 session_len;
    offset += 32; /*skip random*/
    // Read session
    if (bpf_skb_load_bytes(skb, offset, &session_len, sizeof(session_len)) < 0) {
        return TLSTRACKER_UNKNOWN;
    }
    offset += 1 + session_len;
    // Read cipher suite (same position in all versions)
    if (bpf_skb_load_bytes(skb, offset, &tls->cipher_suite, sizeof(tls->cipher_suite)) < 0) {
        return TLSTRACKER_UNKNOWN;
    }
    tls->cipher_suite = bpf_ntohs(tls->cipher_suite);
    offset += 3; // Skip also compression (1B)
    if (handshake_version == 0x0303) {
        // Check extensions to discriminate 1.2 and 1.3
        u16 exts_len;
```

bpf/tls_tracker.h
```diff
@@ -0,0 +1,285 @@
...
+        return TLSTRACKER_NOTLS;
+    }
+
+    struct tls_record rec;
+    u32 offset = (long)l4_hdr - (long)skb->data + len;
+
+    if ((bpf_skb_load_bytes(skb, offset, &rec, sizeof(rec))) < 0) {
+        return TLSTRACKER_NOTLS;
+    }
+    offset += 5;
```
jpinsonneau, line 222: Can you please put a comment to explain this ?

bpf/tls_tracker.h
```diff
@@ -0,0 +1,285 @@
...
+
+    switch (rec.content_type) {
+    case CONTENT_TYPE_HANDSHAKE: {
+        // Handshakes should have the handshake header, except if it's a FINISHED msg, which is encrypted.
+        // In both cases, there should be sufficient data to fill tls_handshake_header; but we can't assume it's valid
+        struct tls_handshake_header handshake;
+        if (bpf_skb_load_bytes(skb, offset, &handshake, sizeof(handshake)) < 0) {
+            return TLSTRACKER_NOTLS;
+        }
+        offset += 4;
```
jpinsonneau, line 241: Same here. A comment would help.

bpf/tls_tracker.h
jotak: Pre-1.2 should already be handled - before 1.2, the TLS version was not provided through an extension, it was in the handshake version field (read from lines 41-46 above).
So it would skip the whole `if (handshake_version == 0x0303)` block and return correctly with the handshake version.

jotak:
thanks @jpinsonneau , comments addressed

jpinsonneau:
LGTM

PR #824, opened 2025-11-03
WIP improve perfs

jpinsonneau:
### 1. Lock-free flow updates (`bpf/flows.c`, `bpf/types.h`)
- Removed `bpf_spin_lock` from the `flow_metrics` structure.
- Replaced with atomic operations:
  - `__sync_fetch_and_add()` for `packets` and `bytes`.
  - Direct writes for other fields (idempotent or acceptable occasional races).

Why: Reduces lock contention in the hot path; updates are safe with atomics.

#### `add_observed_intf()` function (`bpf/flows.c`)
- Unrolled the loop for up to 6 interfaces.
- Direct index comparisons instead of a loop.
- Early exits for common cases (0–3 interfaces).

Why: Removes loop overhead; most flows see 1–2 interfaces.

#### `md_already_exists()` function (`bpf/network_events_monitoring.h`)
- Unrolled the loop for the 4-element array.
- Direct comparisons for all positions.

Why: Eliminates loop overhead in network event checking.

### 3. Early IP filtering (`bpf/flows_filter.h`, `bpf/flows.c`, `bpf/utils.h`)
- Added `early_ip_filter_check()` for IP-only rejection without L4 parsing.
- Split parsing into:
  - `fill_ethhdr_l3only()` — parses L2+L3 only
  - `parse_l4_after_l3()` — parses L4 separately
  - `fill_iphdr_l3only()` / `fill_ip6hdr_l3only()` — L3-only variants

Why: Skips L4 parsing when IP-based filtering can reject packets early, reducing work.

### 4. Memory initialization optimizations (`bpf/flows.c`)
- Replaced `__builtin_memset()` with explicit field initialization.
- Uses designated initializers (`flow_metrics new_flow = { ... }`) and selective initialization.
- Initialize only necessary fields; compiler handles the rest.

Why: Avoids unnecessary zeroing; compiler can optimize better.

### 5. Generated Go code updates
- Updated `pkg/ebpf/bpf_*_bpfel.go` (all architectures) to remove the `Lock` field.
- Updated `pkg/model/record_test.go` to reflect the removed lock field in binary encoding tests.

## Overall impact

These changes target high-frequency paths:
1. Lock-free updates — reduces contention.
2. Loop unrolling — removes loop overhead.
3. Early filtering — skips unnecessary L4 parsing.
4. Better initialization — fewer unnecessary memory operations.

Together, these reduce CPU cycles per packet, which should improve throughput in a high-traffic eBPF flow monitoring agent.

## Dependencies

n/a

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,10 +19,15 @@ COPY go.sum go.sum
 # Build
 RUN CGO_ENABLED=0 GOARCH=$TARGETARCH go build -ldflags "$LDFLAGS" -mod vendor -a -o bin/netobserv-ebpf-agent cmd/netobserv-ebpf-agent.go
 
+# Build collect-kernel-stats utility
+# Note: CGO may be required for cilium/ebpf library interactions with kernel
+RUN GOARCH=$TARGETARCH go build -mod vendor -a -o bin/collect-kernel-stats ./cmd/collect-kernel-stats
+
 # Create final image from minimal + built binary
 FROM --platform=linux/$TARGETARCH registry.access.redhat.com/ubi9/ubi-minimal:9.6-1758184547
 WORKDIR /
 COPY --from=builder /opt/app-root/bin/netobserv-ebpf-agent .
+COPY --from=builder /opt/app-root/bin/collect-kernel-stats /collect-kernel-stats
 USER 65532:65532
 
 ENTRYPOINT ["/netobserv-ebpf-agent"]
--- a/Makefile
+++ b/Makefile
@@ -159,6 +159,11 @@ compile: ## Compile ebpf agent project
 	@echo "### Compiling project"
 	GOARCH=${GOARCH} GOOS=$(GOOS) go build -mod vendor -o bin/netobserv-ebpf-agent cmd/netobserv-ebpf-agent.go
 
+.PHONY: collect-kernel-stats
+collect-kernel-stats: ## Build kernel-space eBPF statistics collector
+	@echo "### Building kernel stats collector"
+	GOARCH=${GOARCH} GOOS=$(GOOS) go build -mod vendor -o bin/collect-kernel-stats ./cmd/collect-kernel-stats
+
 .PHONY: test
 test: ## Test code using go test
 	@echo "### Testing code"
--- a/cmd/collect-kernel-stats.go
+++ b/cmd/collect-kernel-stats.go
@@ -1,0 +1,243 @@
+//go:build linux
+
+package main
+
+import (
+	"encoding/json"
+	"errors"
+	"flag"
+	"fmt"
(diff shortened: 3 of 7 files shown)


jpinsonneau:
Add a python script to compare perfs: https://github.com/netobserv/netobserv-ebpf-agent/pull/824/commits/53d49f86d25f0be1d38691e7c8520f0c4f9c15a7

ebpf_performance_visualization

I was expecting better performances improvments here but it still handle more flows and the ratio is showing improvments in terms of memory.

WDYT @jotak ?
cc @msherif1234 you way me interested too :smile_cat:

jpinsonneau:
Comparing to last 3 runs shows better results: https://github.com/netobserv/netobserv-ebpf-agent/pull/824/commits/dda38e53891a892e4f7f44a33d73b31b3221f021

ebpf_performance_visualization

bpf/flows.c
```diff
@@ -58,48 +58,113 @@
...
-        if (value->observed_intf[i] == if_index) {
-            if (value->observed_direction[i] != direction &&
-                value->observed_direction[i] != OBSERVED_DIRECTION_BOTH) {
-                // Same interface seen on a different direction => mark as both directions
-                value->observed_direction[i] = OBSERVED_DIRECTION_BOTH;
-            }
-            // Interface already seen -> skip
-            return 0;
+
+    // Fast path: unroll loop for small array sizes (most common cases)
```
jotak, line 68: I think we must measure that to see how much it improves CPU. The downside I see is that the code is less intuitive / readable, and also it's error prone if we decide to increase MAX_OBSERVED_INTERFACES (we'd need to add new "unrolled" blocks, which can easily be missed)
But optimizations often come with tradeoff so that might be ok, depending on the measured improvement

bpf/flows.c
```diff
@@ -183,18 +273,20 @@ static inline int flow_monitor(struct __sk_buff *skb, u8 direction) {
...
+            .packets = 1,
+            .bytes = len,
+            .eth_protocol = eth_protocol,
+            .start_mono_time_ts = pkt.current_ts,
+            .end_mono_time_ts = pkt.current_ts,
+            .flags = pkt.flags,
+            .dscp = pkt.dscp,
+            .sampling = flow_sampling,
+            .nb_observed_intf = 0 // Explicitly zero for clarity
+        };
```
jotak, line 289: we used to do that previously, and switched to individual assignments, iirc @msherif1234 found cases where that didn't work as intended, but can't remember what exactly. @msherif1234 do you remember?

jpinsonneau:
Closing as outdated. Will reopen if I have opportunity to spend more time on this in future.

PR #879, opened 2026-02-02
NETOBSERV-2589 fix bpfman SSL tracking

jpinsonneau:
## Description

Load SSL data pinned map only when SSL Tracking is enabled to avoid error:
```
time="2026-01-29T22:19:26Z" level=info msg="BPFManager mode: loading SSL data event pinned maps" component=ebpf.FlowFetcher
time="2026-01-29T22:19:26Z" level=fatal msg="can't instantiate NetObserv eBPF Agent" error="failed to load /run/netobserv/maps/ssl_data_event_map: no such file or directory"
```

## Dependencies

n/a

## Checklist

### DIFF
--- a/pkg/tracer/tracer.go
+++ b/pkg/tracer/tracer.go
@@ -410,12 +410,23 @@ func NewFlowFetcher(cfg *FlowFetcherConfig, m *metrics.Metrics) (*FlowFetcher, e
 		if err != nil {
 			return nil, fmt.Errorf("failed to load %s: %w", mPath, err)
 		}
-		log.Infof("BPFManager mode: loading SSL data event pinned maps")
-		mPath = path.Join(pinDir, sslDataEventMap)
-		objects.BpfMaps.SslDataEventMap, err = cilium.LoadPinnedMap(mPath, opts)
-		if err != nil {
-			return nil, fmt.Errorf("failed to load %s: %w", mPath, err)
+
+		// Only load SSL map if OpenSSL tracking is enabled
+		if cfg.EnableOpenSSLTracking {
+			log.Infof("BPFManager mode: loading SSL data event pinned maps")
+			mPath = path.Join(pinDir, sslDataEventMap)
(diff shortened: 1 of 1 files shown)


msherif1234:
Hi @jpinsonneau we don't do the same check for other maps have we update the bpfmap operator part with ssl ? I suspect the error because github.com/netobserv/netobserv-ebpf-agent/pkg/maps is not latest ?

oliver-smakal:
Hi @jpinsonneau, Testing it using the ebpf-agent pre-merge image works for me with the 'main' netobserv version , but not with the 1.11 one. Is this expected?

jpinsonneau:
> Hi @jpinsonneau we don't do the same check for other maps have we update the bpfmap operator part with ssl ? I suspect the error because github.com/netobserv/netobserv-ebpf-agent/pkg/maps is not latest ?

Oh indeed I will bump the operator in that case :ok_hand: 
Thanks for the tips

Edit: actually it's already there -> https://github.com/netobserv/network-observability-operator/blob/release-1.11/vendor/github.com/netobserv/netobserv-ebpf-agent/pkg/maps/maps.go#L19 :thinking: 

I think it will require the program to be loaded which is not the case since the feature doesn't exists yet on operator side: https://github.com/netobserv/network-observability-operator/pull/2398/changes#diff-6cfe17bbb1db730676598586bd4e1472370cb5f2d84f90a5835af1dd360308e4R212-R214

Since we are in a release phase, that fix would be easier to integrate for now. We can still rollback it after the release.

oliver-smakal:
> Hi @jpinsonneau, Testing it using the ebpf-agent pre-merge image works for me with the 'main' netobserv version , but not with the 1.11 one. Is this expected?

My bad, upon retrying with 1.11 one more time, it works, I must have made some kind of mistake before.

jotak:
+1 to @jpinsonneau approach; we should actually do the same thing for all maps, so it's less coupled with the operator code.

PR #976, opened 2026-05-11
TLS: surface known versions in priority in ClientHello

jotak:
Fixes an issue where TLS version shows weird names such as "0xDADA"
See comment here https://redhat.atlassian.net/browse/NETOBSERV-2608?focusedCommentId=16942064

We now prioritize showing well known versions over unknown ones

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TLS version detection and validation during handshakes.
  * Enhanced negotiation logic to prefer known protocol versions while still selecting the highest compatible version.
  * Better handling of multiple TLS version candidates for increased connection reliability and compatibility.

[Review Change Stack](https://app.coderabbit.ai/change-stack/netobserv/netobserv-ebpf-agent/pull/976)


Amoghrd:
/jira [NETOBSERV-2608](https://redhat.atlassian.net/browse/NETOBSERV-2608)

bpf/tls_tracker.h
```diff
@@ -104,7 +107,12 @@ static inline int tls_read_client_hello(struct __sk_buff *skb, u32 offset, tls_i
                         return TLSTRACKER_UNKNOWN;
                     }
                     version = bpf_ntohs(version);
-                    if (version > tls->hello_version) {
+                    // Favor either known versions over unknown (e.g. 0x0304 over 0xDADA) or the higher one (e.g. 0x0304 over 0x0303)
+                    bool is_known = IS_KNOWN_VERSION(version);
+                    if (!current_is_known && is_known) {
+                        tls->hello_version = version;
+                        current_is_known = true;
+                    } else if (current_is_known == is_known && version > tls->hello_version) {
```
msherif1234, line 116: if current and is_known both `false` can't be end up with weired version in tls->hello_version ? why you are doing 5 iterations here ?

jotak: > if current and is_known both false can't be end up with weired version in tls->hello_version ?

yes it can, but I don't think it's wrong? IMO as an observability tool, it must report what it sees, even the weird things

> why you are doing 5 iterations here ?

we need to bound the loop; 5 is arbitrary, it could be 10 or 20; we just need to pick something.

Amoghrd:
Looks good!

PR #994, opened 2026-06-19
NETOBSERV-2273: support multiple DNS tracking ports via comma-separated list

OlivierCazade:
## Description

Update DNS_TRACKING_PORT to accept comma-separated port values while maintaining backward compatibility with single port configuration.

Changes:
- Add support for up to 8 DNS ports (MAX_DNS_PORTS)
- Replace single dns_port with dns_ports array in eBPF config
- Remove hardcoded port 53 check in track_dns_packet()
- Add is_dns_port() helper with bounded loop for eBPF verifier
- Update config to parse DNS_TRACKING_PORT as comma-separated string
- Update documentation with multi-port usage examples

Examples:
- DNS_TRACKING_PORT=53 (single port, backward compatible)
- DNS_TRACKING_PORT=53,5353,8053 (multiple ports)

## Dependencies

n/a

## Checklist

* [ ] Does the changes in PR need specific configuration or environment set up for testing?
   * [ ]  if so please describe it in PR description.
* [X] I have added thorough unit tests for the change.

## Summary by CodeRabbit

* **New Features**
  * DNS tracking now supports monitoring multiple DNS ports simultaneously (up to 8) via `DNS_TRACKING_PORT` as a comma-separated list (default: `53`, e.g., `53,5053`).
* **Bug Fixes**
  * DNS packet filtering now matches traffic when either the source or destination port is in the configured port list (instead of a single port).
* **Documentation**
  * Updated `DNS_TRACKING_PORT` docs with multi-port format and max supported count.
* **Tests**
  * Added tests covering parsing behavior, whitespace trimming, truncation to 8 ports, invalid/empty entries, edge ports, and config disabled behavior.


bpf/dns_tracker.h
```diff
@@ -65,11 +64,24 @@ static __always_inline u8 calc_dns_header_offset(pkt_info *pkt, void *data_end)
     return len;
 }
 
+static __always_inline bool is_dns_port(u16 port) {
+    // Bounded loop for eBPF verifier
+    #pragma unroll
```
msherif1234: no need to use those pragma the compliers take care of that we no longer need to use them

bpf/dns_tracker.h
```diff
@@ -65,11 +64,24 @@ static __always_inline u8 calc_dns_header_offset(pkt_info *pkt, void *data_end)
...
+    for (u8 i = 0; i < MAX_DNS_PORTS; i++) {
+        if (i >= dns_ports_count) {
+            break;
+        }
+        if (port == dns_ports[i]) {
```
msherif1234, line 73: have u checked out using global array map table to pass in this configuration instead of using global variables array ?

bpf/dns_tracker.h
OlivierCazade: Done, thanks

bpf/dns_tracker.h
OlivierCazade: To be honest I am not sure to fully understand the differences between both. Could you elaborate ?

As far as I can tell, the project is using global variables when needed.

msherif1234: we use global as config knobs but I think for passing an array which operator can change runtime its better to store the list as an array map from user space and let the kernel read it

OlivierCazade: This config comes from env, the operator can not change it without recreating the pod.

msherif1234: how many ports this can be ? as global u will always burn that space regardless if DNS is enabled or not, but if its few ports then that might not be a concern ?

OlivierCazade: For now, there is a hard limit of 8 different ports.

pkg/config/config.go
```diff
@@ -234,9 +234,9 @@ type Agent struct {
 	EnablePktDrops bool `env:"ENABLE_PKT_DROPS" envDefault:"false"`
 	// EnableDNSTracking enable DNS tracking eBPF hook to track dns query/response flows
 	EnableDNSTracking bool `env:"ENABLE_DNS_TRACKING" envDefault:"false"`
-	// DNSTrackingPort used to define which port the DNS service is mapped to at the pod level,
-	// so we can track DNS at the pod level
-	DNSTrackingPort uint16 `env:"DNS_TRACKING_PORT" envDefault:"53"`
+	// DNSTrackingPorts used to define which ports the DNS service is mapped to at the pod level,
+	// so we can track DNS at the pod level. Comma-separated list of ports (e.g., "53,5353,8053")
+	DNSTrackingPorts string `env:"DNS_TRACKING_PORT" envDefault:"53"`
```
msherif1234: wouldn't it be cleaner if we define it as 
```
DNSTrackingPorts []uint16
```

OlivierCazade: Changed, thanks!

jotak (dismissed):
LGTM

memodi:
@OlivierCazade please rebase.

PR #1040, opened 2026-09-02
tune coderabbit to ignore vendor files

jotak:
I think the "**" is needed; e.g. see https://github.com/netobserv/flowlogs-pipeline/pull/1329 where it ignored the review due to too many vendor files.
If confirmed we'll need to apply it everywhere.
cc @memodi 

## Summary by CodeRabbit

* **Chores**
  * Updated review path filtering to exclude all files and subdirectories within the vendor directory.


memodi:
/lgtm

thanks @jotak

PR #1048, opened 2026-09-10
perf(flow): batch-allocate records during eviction

awprice:
## Description

Reduce heap allocations when the map tracer and userspace accounter evict a batch of flows. Both paths now allocate the `Record` values, first-interface storage, and record pointers in contiguous batches instead of allocating a `Record` and interface slice for every flow.

`NewRecordInto` initializes a caller-provided record while keeping `NewRecord` available for existing callers. Capacity-limited interface subslices keep each record's first interface isolated when a flow has additional observed interfaces. A unit test covers that case.

### Benchmark

Benchmarked the parent commit and this commit on an Apple M4 Pro with Darwin/arm64 and Go 1.25.7. Each result is based on six runs:

```sh
GOMAXPROCS=8 go test -mod=vendor -run '^$' -bench '^BenchmarkEvictFlows$' -benchmem -count 6 ./pkg/flow
```

| Flows per eviction | Base B/op | New B/op | Change | Base allocs/op | New allocs/op | Change |
| ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 1k | 506.6 KiB | 488.2 KiB | -3.63% | 2,019 | 11 | -99.46% |
| 10k | 4.668 MiB | 4.305 MiB | -7.77% | 20,054 | 39 | -99.81% |
| 100k | 44.76 MiB | 39.72 MiB | -11.25% | 200,288 | 263 | -99.87% |

The allocation and byte differences were statistically significant at each size (`p=0.002`, `n=6`). The 10k case also reduced median runtime from 1.381 ms to 0.947 ms (`-31.47%`, `p=0.002`). Runtime results at 1k and 100k were too noisy to draw a conclusion.

### Testing

```sh
go test -mod=vendor ./pkg/model ./pkg/flow
go test -mod=vendor -race ./pkg/model ./pkg/flow
```

No configuration or environment changes are needed to test this change.

## Dependencies

n/a

## Checklist

* [ ] Does the changes in PR need specific configuration or environment set up for testing?
   * [ ]  if so please describe it in PR description.
* [x] I have added thorough unit tests for the change.
  * [x] Regression tests only (e.g. refactoring with no user-facing change).

## Summary by CodeRabbit

- **Performance**
  - Improved record and flow processing efficiency by reducing memory allocations.
  - Improved efficiency across eviction operations while preserving existing behavior, metrics, channel output, and processing results.
  - Expanded performance coverage across flows with one or multiple interfaces.

- **Bug Fixes**
  - Prevented interface data from being overwritten when record storage is reused.
  - Preserved accurate interface information while processing reused records.


jotak:
thanks again @awprice , let's check the impact on our performance tests

pkg/flow/tracer_map.go
```diff
@@ -118,15 +117,25 @@ func (m *MapTracer) evictFlows(ctx context.Context, forceGC bool, forwardFlows c
 			mtlog.Tracef("GetInterfaceUDNS map: %v", udnCache)
 		}
 	}
+	// Records are forwarded and released as a batch, so allocate their values and
+	// common first-interface storage as contiguous blocks.
+	recordsBacking := make([]model.Record, len(flows))
+	interfacesBacking := make([]model.IntfDirUdn, len(flows))
```
jotak, line 123: This seems to be optimized for 1 interface per flow, am I getting it right? Is it how it works for you (with Cilium I guess) ?

In ovn's use case it's common to see several interfaces per flow, so this optimization seems less relevant. But if I'm correct, it's still net positive, with `nbInterfaces - 1` allocs per flow

pkg/model/record.go
```diff
@@ -101,12 +119,12 @@ func NewRecord(
 	if metrics.DirectionFirstSeen == 0 {
 		lMAC = metrics.DstMac
 	}
-	record.Interfaces = []IntfDirUdn{NewIntfDirUdn(interfaceNamer(int(metrics.IfIndexFirstSeen), lMAC),
+	dst.Interfaces = append(interfaces, NewIntfDirUdn(interfaceNamer(int(metrics.IfIndexFirstSeen), lMAC),
```
jotak, line 122: please fix the linter error or add `// nolint:gocritic` with a short explanation

pkg/flow/tracer_map.go
jotak: I ran the benchmark with an extra interface: as expected the improvement is more nuanced then, although still very positive:

before this PR:
```
BenchmarkEvictFlows/1k-14 	    1274	    948529 ns/op	  598714 B/op	    3019 allocs/op
BenchmarkEvictFlows/10k-14         	     144	   8602642 ns/op	 5695119 B/op	   30054 allocs/op
BenchmarkEvictFlows/100k-14        	      12	 101158658 ns/op	54933352 B/op	  300288 allocs/op
```

with this PR:
```
BenchmarkEvictFlows/1k-14 	    1333	    841693 ns/op	  579907 B/op	    1011 allocs/op
BenchmarkEvictFlows/10k-14         	     145	   7797820 ns/op	 5314547 B/op	   10039 allocs/op
BenchmarkEvictFlows/100k-14        	      13	  84589409 ns/op	49653485 B/op	  100263 allocs/op
```

@awprice could you add a variant of your benchmark  that uses more interfaces (2 or 3, I would say) so it doesn't become a blind spot?

awprice: Yep, that's right. We're using Cilium, and in our environment we see one interface per flow.

I've added a three-interface benchmark alongside the existing one-interface case. I reran both against the parent commit using six runs per case:

| Interfaces | Flows | Time/op | B/op | Allocs/op |
| ---: | ---: | ---: | ---: | ---: |
| 1 | 1k | 161.2µs → 120.1µs (-25.51%) | 506.6KiB → 488.2KiB (-3.63%) | 2,019 → 11 (-99.46%) |
| 1 | 10k | 1.476ms → 896.6µs (-39.24%) | 4.668MiB → 4.305MiB (-7.77%) | 20,054 → 39 (-99.81%) |
| 1 | 100k | 17.31ms → 13.07ms (-24.47%) | 44.76MiB → 39.72MiB (-11.25%) | 200,288 → 263 (-99.87%) |
| 3 | 1k | 270.0µs → 224.0µs (-17.04%) | 740.9KiB → 722.6KiB (-2.48%) | 4,019 → 2,011 (-49.96%) |
| 3 | 10k | 2.534ms → 1.877ms (-25.95%) | 6.957MiB → 6.594MiB (-5.22%) | 40,054 → 20,039 (-49.97%) |
| 3 | 100k | 25.46ms → 19.47ms (-23.54%) | 67.65MiB → 62.61MiB (-7.44%) | 400,288 → 200,263 (-49.97%) |

The three-interface case keeps the expected extra allocations, but the change still removes about half of them.

jotak: thanks!

jotak (approved):

jotak:
@awprice not related to your PR but I have a question or two and pinged you on CNCF slack, if you don't mind to take a look

